PR #4, opened 2017-02-27
added use cases as discussed in scripting call

h0ru5:
results of scripting call on 2017-02-27

### DIFF
--- a/index.html
+++ b/index.html
@@ -251,7 +251,8 @@
   <p>
     The following scripting use cases are covered in this specification:
     <ul>
-      <li>Discover <a>Thing</a>s by filters defined on <a>Thing Description</a>s.</li>
+      <li>Discover <a>Thing</a>s by filters defined on <a>Thing Description</a>s for client-side access</li>
+      <li>Discover <a>Thing</a>s by semantic filters for client-side access</li>
       <li><a>Consume a TD</a> of a remote <a>Thing</a>.</li>
       <li>On a consumed <a>Thing</a>,
         <ul>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -262,14 +263,19 @@
...
         </ul>
       </li>
+      <li>Retrieve a thing that is exposed by the local runtime</li>
       <li>Create and expose a local <a>Thing</a> based on a <a>Thing Description</a>.</li>
       <li>Programmatically create and expose a local <a>Thing</a>. This may include the following operations:
         <ul>
           <li>Add a property definition to the <a>Thing</a>.</li>
           <li>Add an event definition to the <a>Thing</a>.</li>
           <li>Add an action definition to the <a>Thing</a>.</li>
+          <li>Attach semanic information to an Action</li>
```
zolkis: s/semanic/semantic

index.html
```diff
@@ -262,14 +263,19 @@
...
+          <li>Attach semanic information to a Property</li>
```
zolkis: same

index.html
```diff
@@ -262,14 +263,19 @@
...
+          <li>Attach semanic information to an Event</li>
```
zolkis: same

index.html
```diff
@@ -262,14 +263,19 @@
...
           <li>Start the exposed <a>Thing</a> in order to process external requests.</li>
-          <li>Handle an external request to retrieve the <a>Thing Description</a>.</li>
+          <li>Modify the handling of an external request to retrieve the <a>Thing Description</a>.</li>
```
zolkis: some reformulation would be needed; does it mean a custom handler for the external request to retrieve the TD, or a piece of code that can act upon the Thing and TD before the implementation generates the response?

index.html
```diff
@@ -262,14 +263,19 @@
...
+          <li>React on a property change</li>
```
zolkis: perhaps would be more accurate to say "Handle an external request to change a property" (called by the implementation before the property is changed)?

h0ru5: so far we have discussed it as it in the way a $watch works in Angular 1. That is to say, the callback*s* get called upon updates with newValue and oldValue as parameter.
This enables to have several callbacks reacting on a change, but does not enable modifying or denying the change (dumb setter).
Use case is e.g. I have an RGB light with 3 colors, and when someone changes the value of RED, I change the physical ledstate (and possibly do some side-effect).

I could think of a pattern like the "handler gets called with (newValue, oldValue) and returns a value that is assigned to the property" but that would limit it to one handler IMHO.

index.html
h0ru5: I was actually thinking in the latter direction - cases might be that i want just part of the TD to go out or filter it to one single protocol (the one from the request) etc.

zolkis:
LGTM, merging.

Issue #6, opened 2017-03-05
Add use cases


knimura:
Wouldn't it be necessary to have the following use cases?

[ConsumedThing]
- Currently only mentions consume remote thing.
  use case for local string/object?
- Use case for remove all listeners.

[ExposedThing]
- Use case for remove all listeners.
- Use case for unregister a thing.
- Use case for start.

[Programmable]
- use case for add a handler which is called when an action is invoked.
  this is different from use case for addAction which only defines name and parameter/return type for action.
- use case for clear ActionHander
- use case for add a handler which is called when a property value is updated
- use case for multiple add PropertyChangeListeners
- use case for remove a handler which is called when a property is updated
- use case for remove all handlers which are called when a property is updated

zolkis:
For `ExposedThing`:
- start and stop is already there in the use cases.
- register and unregister seems to be missing, we need to add.
- removing all listeners means removing other clients from observing a Thing. This is very invasive and until we have a solid security model, IMHO should not be encouraged. What is the application use case for this scripting use case?

For `ConsumedThing`:
- IMHO it should make no difference in the API if the consumed Thing is local or remote; it is identified by a URI. It's the bindings/implementation that should care about local optimization, IMHO we should not create a local and remote API for consumed and exposed things. Arguably, if there is reason for local APIs, they should go in a different spec, and appeal to WoT bindings/implementations.

Other
- what is the application use case for action hooks? observing/deflecting by other clients when an action is triggered by some other client? So every client can register an action hook? If so, IMHO we need to think it over find a better abstraction for this (and the pattern in general, not only including actions). 
If this is for the local script, IMHO hooks are not needed, since it can be done locally in script.
So, code examples would be helpful in order to clarify what do we want to achieve here.

zolkis:
Should we add flags to an `ExposedThing` whether they are meant to be discoverable, observable, etc? (policies)

zolkis:
As mentioned [here](https://github.com/w3c/wot-scripting-api/pull/7#issuecomment-285650009) I still don't see the value of hooks on property changes and action invocations.

Observing property changes and action invocations should be part of the client API.
On the server, since the hooks would be defined by the same script that defines the server Thing, that logic could go directly in the request handlers.

Could someone provide code examples for why server side hooks are needed?

knimura:
Latest use cases seem fine.
Lifecycle of script could be added if we agreed on.

h0ru5:
I have one little detail:
[...]

Issue #3, opened 2017-02-27
Use constructors for WoT objects


zolkis:
Since WoT is going to be implemented on different types of runtimes (see also #2), 
and since the *root* [1] WoT object has no state,
arguably it would be simpler to expose the WoT object with a constructor (internally implementations could still use a singleton).
Also, for `ExposedThing` and `ConsumedThing` we could provide constructors instead of factories. That would be more aligned with ECMAScript best practices (e.g. an `ExposedThing` object could be created for testing purposes, shaped locally, then exposed).

Currently the interface uses factories:
```javascript
interface WoT {
    Promise<void>          discover(ThingFilter filter,
                                    ThingDiscoveryCallback onfound);

    Promise<ConsumedThing> consumeDescription(object thingDescription);
    Promise<ConsumedThing> consumeDescriptionUri(DOMString thingDescriptionURI);

    Promise<ExposedThing>  createThing(DOMString name);
    Promise<ExposedThing>  createFromDescription(object thingDescription);
    Promise<ExposedThing>  createFromDescriptionUri(DOMString thingDescriptionURI);
};
```

It could be changed into:

```javascript
[Constructor()]
interface WoT {
    Promise<void>          discover(ThingFilter filter,
                                    ThingDiscoveryCallback onfound);
};

[Constructor(DOMString name)]
[Constructor(object ThingDescription)]
[Constructor(DOMString thingDescriptionURI)]
interface ExposedThing {
  // ... 
 Promise<void> expose();  // register with the WoT network
};

[Constructor(object ThingDescription)]
[Constructor(DOMString thingDescriptionURI)]
interface ConsumedThing {
  // ... 
  Promise<void> fetch();  // retrieve/validate the Thing
};
 
```
[1] edited: s/Wot object/root WoT object

draggett:
> and since the WoT object has no state

what makes you assume that?  A binary switch has state. Perhaps you are using state in a specialised sense, if so could you please explain.

zolkis:
Misunderstanding. I meant the root WoT object. It doesn't have properties, only functions.

knimura:
Wouldn't it be necessary to have separate interfaces for the "object creation" and "exposed"?

zolkis:
They would be separated: `new` creates the object (which is not  bound to the underlying platform yet), it's just a JS object. Then, for `ExposedThing`, invoke `exposed()` to make it a WoT Thing (register etc).
[...]

PR #7, opened 2017-03-06
Add use cases. Refactor WebIDL.

zolkis:
WebIDL was updated based on comments in #3.
Added some of the use cases mentioned in #6.

### DIFF
--- a/index.html
+++ b/index.html
@@ -259,8 +259,20 @@
           <li>Get a property value.</li>
           <li>Set a property value.</li>
           <li>Invoke an action.</li>
-          <li>Add a listener to an event.</li>
-          <li>Remove a listener from an event.</li>
+          <li>
+            Observe events.
+            <ul>
+              <li>Add a listener to an event.</li>
+              <li>Remove a listener from an event.</li>
+            </ul>
+            Default events are the following:
+            <ul>
+              <li>A property has changed.</li>
+              <li>An action has been invoked.</li>
+              <li>Other Things have added an event listener.</li>
+              <li><a>Thing Description</a> has changed, i.e. properties, events or actions have been defined, removed, or the definition has changed.</li>
+            </ul>
+          </li>
         </ul>
       </li>
       <li>Retrieve a thing that is exposed by the local runtime</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -267,20 +267,23 @@
...
           <li>Add a property definition to the <a>Thing</a>.</li>
           <li>Add an event definition to the <a>Thing</a>.</li>
           <li>Add an action definition to the <a>Thing</a>.</li>
-          <li>Attach semantic information to an Action</li>
-          <li>Attach semantic information to a Property</li>
-          <li>Attach semantic information to an Event</li>
-          <li>Start the exposed <a>Thing</a> in order to process external requests.</li>
-          <li>Modify the handling of an external request to retrieve the <a>Thing Description</a>.</li>
-          <li>React on a property change</li>
+          <li>Register an action triggered by external retrieve requests on the <a>Thing Description</a>.</li>
```
h0ru5: to avoid confusion with the interacion pattern "action" (of a thing) , could we say "a callback" or "logic"?

h0ru5: to avoid confusion with the interacion pattern "action" (of a thing) , could we say "a callback" or "logic"?

zolkis:
What about the following:
- for request handlers we use single callback (to handle requests like retrieve TD, run Action, add Event listener, etc)
- for hooks we use built-in events, like `on("actionrequest")`, `on("propertychangerequest") to which multiple listeners can be attached. In the use case, we can use the term "observe", like "Observe requests for invoking an action", and "Observe a request for property change". The effect is that after a property is changed or an action is invoked, listeners are notified in order to update their business logic.

However, I still fail the understand the use case for the hooks. These hooks are only used by the same script that defines the server Thing. Why cannot the logic that goes into the hooks go directly into the request handlers?

zolkis:
I have reformulated the use cases and moved all observation use cases to consumed Thing. The misunderstanding probably came from the fact that ExposedThing also implements ConsumedThing and in the Current Practices they are just dumped together.

zolkis:
@knimura @h0ru5 could you review again, please

zolkis:
Added request handlers to `ExposedThing` and consolidated use of dictionaries for properties, actions and events.

zolkis:
Based on [this comment](https://github.com/w3c/wot-scripting-api/issues/6#issuecomment-286037138) I am merging for now. If there is feedback, let's raise them later.

Issue #18, opened 2017-04-03
Adding readonly properties through ExposedThing interface


zolkis:
_From @retamar on March 3, 2017 8:1_

There are any way to add a readonly property using the scripting API through ExposedThing interface?  I see that in TD we have the "writable" property but, there are no way to set it's value through ExposedThing. It seems that is not possible to assign any other property (or any interaction) metadata using the scripting AP such as stability or semantic data such as "units".

_Copied from original issue: w3c/wot#310_

zolkis:
_From @h0ru5 on April 3, 2017 11:37_

decision to move over to new  repo

zolkis:
Moved the issue over from the WoT IG repo.

h0ru5:
The most basic idea would be to add a boolean field ``writeable`` to the ``ThingProperty`` dictionary
I'd suggest defaulting to ``true``

zolkis:
Fixed in #20.

Issue #14, opened 2017-03-27
``ExposedThing`` and ``DynamicThing``


h0ru5:
despite the naming should be reconsidered, we have a general question to discuss about  regarding the ``ExposedThing`` and ``DynamicThing``, where one has static interface during runtime, while the other is modifiable by methods like ``addProperty`` or ``removeAction`` 

This issue is  forked off from the discussion on #6 and also relates to the general contract between Things, Scripts and WoT Runtime as discussed in #2

zolkis:
If the difference is just a policy for updating the TD, it could be exposed as a policy property (provided during construction either implicitly by the runtime configuration or explicitly by the script). Then this policy could possibly be encapsulated by the runtime... 
The runtime would decide if a given TD update request can be fulfilled, otherwise return an error. It would simplify Things.

h0ru5:
so from the viewpoint of a script dev: 

the runtime would always offer all methods but give an error if I try to change an "immutable" thing (e.g. an ``ExposedThing`` which is not a ``DynamicThing``). However, I can check that condition on with property in advance.

Right?

zolkis:
Yes, that's the idea.

knimura:
Let me confirm:
1. The ability of runtime for ExposedThing or DynamicThing is reflected to a policy definition in a property when generating a TD.
2. A script that tries to update the runtime by e.g. addProperty can check that condition in advance by reading the property regardless whether the script is located inside the runtime or outside of it.
3. The runtime will be updated if it has the ability of DynamicThing. If not, and still someone ask the request, the runtime will return an error.
Correct?

zolkis:
Yes, that's my understanding. What we don't know yet is how exactly will we expose and handle policy information - but we can start with a string or dictionary property.

zolkis:
Of course we could keep the names, e.g. Exposed Thing (not ExposedThing), but in the API we'd have one type of object. The rationale for this would be that 
- It's more generic and can cover future Thing classifications better. When we can parametrize rather than sub-class, let's do that.
- The policy _may_ be changed in runtime (or may not). This way it's more flexible. With ExposedThing the policy would be built in. Usually we should strive to build in mechanisms, not policies, and keep policies external.

zolkis:
Related issue: https://github.com/w3c/wot/issues/310

h0ru5:
my suggestion would be to reflect this policy as a property ``configurable`` of ``ExposedThing`` with the same semantics as 

[...]

PR #20, opened 2017-04-07
adding readOnly entry for Property initializer

h0ru5:
addressing #18 
defaulting to writeable.
using default false value according to [WebIDL recommendation](https://www.w3.org/TR/WebIDL-1/#dfn-dictionary-member-default-value), therefore readOnly instead of writeable.

### DIFF
--- a/index.html
+++ b/index.html
@@ -472,7 +472,7 @@
       object contentType;
       SemanticType[] semanticTypes;
       any value;
-      boolean readOnly = false;
+      boolean writable = false;
     };
 
     dictionary ThingEvent {
--- a/index.html
+++ b/index.html
@@ -472,6 +472,7 @@
       object contentType;
       SemanticType[] semanticTypes;
       any value;
+      boolean readOnly = false;
     };
 
     dictionary ThingEvent {


zolkis:
IMHO we should use the descriptor names for defining JS object properties - at least to be aligned with JS and TypeScript. So, 
`boolean writeable` 
`boolean enumerable`
`boolean configurable` (type may be changed and property may be deleted)

By default all of these is `false` in JS. 
For WoT properties the default may be `true` (limit access explicitly and on need).

index.html
```diff
@@ -472,6 +472,7 @@
       object contentType;
       SemanticType[] semanticTypes;
       any value;
+      boolean readOnly = false;
```
zolkis: I'd suggest `boolean writeable = true;` instead (to be consistent with JS property descriptors).

h0ru5: I'll do writeable [spelling according to ECMA] and default to false

h0ru5: Re-checked ECMAscript convention and will put it to ``writable = false``

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Writable_attribute

@sebastiankb we should do the same in the TD

zolkis: Yes, +1

zolkis:
+1 
please squash :)

h0ru5:
I updated ``writable`` (can the value be changed?). I'd suggest to have ``configurable`` (can the interface be modified?) on the Thing level, as we discuss in #14

Issue #17, opened 2017-04-03
Refactor Discovery API


zolkis:
Based on #16 (stop a discovery), and earlier issues in the WoT IG,
https://github.com/w3c/wot/issues/233 (discover Things from the same Runtime).

Changes requested:
- Use cancelable Promises or Observables as return type
- Use scope for discovery (local, directory, global)
- Define an algorithm on how to "stop" open ended protocol requests for discovery (e.g. broadcasts).

zolkis:
Addressed by #21.

Issue #16, opened 2017-04-03
There are no way to stop thing discovering using ThingFilter + ThingCallback


zolkis:
_From @retamar on March 6, 2017 11:37_

In an execution environment where thing discovery is a long time running operation, if a device needs to find only one thing in a given type, there are no way to stop the discovery operation when first thing was found.
Using ThingDiscoveyCallback you can just ignore new discovered thing, but the lookup operations are still executing.

_Copied from original issue: w3c/wot#311_

zolkis:
_From @h0ru5 on April 3, 2017 11:35_

we will this issue to the scripting repo

zolkis:
Moved the issue over from the WoT IG repo.

zolkis:
Fixed in #21.

Issue #5, opened 2017-02-28
Use Observables instead of callbacks?


zolkis:
As described [here](https://github.com/w3c/wot/blob/master/scripting/open-issues.md#asynchronous-programming-guidelines) and agreed on the face to face meeting,

> Promises should be used for asynchronous operations with one-shot result (either success or fail), and callbacks (later Observables) should be used for asynchronous operations with multiple-shot results (e.g. open ended protocol requests such as discovery).

This issue is for tracking and discussing the WoT use case for [Observables](https://github.com/tc39/proposal-observable).

zolkis:
Use case: return an `Observable` object for open ended [cancelable] asynchronous operations that provides controls over that operation:
- cancel the operation
- attach a callback for handling streamed or partial results
- attach a callback for handling the end of the operation
- attach an error handler callback.

For example:
```javascript
let subscription = asyncOperation(parameters).subscribe({
    next(val) { console.log("Received next result: " + val); },
    error(err) { console.log("Received an error: " + err); },
    complete() { console.log("Operation complete"); }
});
// ...
subscription.cancel();
```
Alternatively, we could say
```javascript
let subscription = asyncOperation(parameters)
    .on("next", val => { console.log("Received next result: " + val); })
    .on("error", function(err) { console.log("Received an error: " + err); })
    .on("complete", () => { console.log("Operation complete"); })
;
```
Or, 
```javascript
let subscription = asyncOperation(parameters)
    .next(val => { console.log("Received next result: " + val); })
    .complete( () => { console.log("Operation complete"); })
    .catch(err => { console.log("Received an error: " + err); });
;
```

zolkis:
This would be used mainly with discovery operations.

zolkis:
We could do a polyfill for the above options, or continue to use callbacks for discovery, but without the ability to cancel discovery, or to know when discovery has ended (when these are supported by the underlying protocols).

h0ru5:
I am +1 on observables for discovery (and maybe also subscription).

I would stick to the current ECMA discussion. I don't know if there is a polyfill already defined, it seems Angular2 used RxJS, which could be an option.

h0ru5:
Conclusion on call 27-03-2017 is to go for a subset of current ECMA discussion, risking a breakage but  betting on promising candidate.
we check the proposal and decide.

h0ru5:
Link to proposal: https://github.com/tc39/proposal-observable

zolkis:
[...]

Issue #2, opened 2017-02-27
Define UA/runtime for WoT


zolkis:
A User Agent in a browser is a runtime that acts on behalf the user and is part of the security model that is based on user approval.

It is difficult to define a *user* for WoT.

Should we use a specific term for `WoT Runtime`, or `WoT UA` or something else?

Temporary formulation:

> A WoT runtime or simply Runtime is defined as a software stack that manages the lifecycle of WoT application scripts. In the case of ECMAScript and TypeScript scripts, it consists of an ECMAScript interpreter, script lifecycle management, and an operating system API that provides access to local and remote resources. A WoT Runtime can be modeled as a virtual machine in the sense that it should be completely isolated from other execution environments on memory address space, storage address space (file system), network namespace, etc. A runtime may be for instance a Node.js environment, an embedded runtime based on a small JavaScript engine, or provided by a browser.
```

knimura:
As you described, assuming a WoT Runtime is modeled as a VM.
It can say that WoT Runtime has three types: WoT Client / WoT Server or WoT Servient on VM. Then each of WoT Runtime can be realized by Node.js.
Especially regarding WoT Client, it can be realized by a browser (UA) extension as well.
Since it just a choice of the mean of WoT Runtime, I think the term of “WoT UA” does not need to define.

zolkis:
I agree. Let's use the full term `WoT Runtime` in the spec until a better term is advised.

h0ru5:
As mentioned in the call on 06-03-2017 we should clarify also the relation of scripts, runtimes and things

zolkis:
IMHO we have 2 options for Runtime vs Thing vs Scripts modeling.

1. A Thing is associated with its own runtime and interacts with other Things only via the WoT interfaces.  Local physical interfaces are encapsulated by bindings. Bindings/implementations may take advantage of local optimizations, but Thing addressing and interfaces don't change based on locality. A script run on a Thing can access other Things based on this standard interface. Script collation/combination is possible following distributed programming paradigms.

[...]

PR #21, opened 2017-04-10
Update WoT runtime, discovery, ExposedThing, ConsumedThing.

zolkis:
Update the spec on text relevant on the following topics:
- WoT Runtime, 
- WoT object
- security considerations
- script handling
- discovery.
Issues addressed: #2  #5  #16 #17 #22 


zolkis:
Based on discussion in #22 , removed the remote Thing management (subject to further debate).

index.html
```diff
@@ -145,19 +142,26 @@
...
     <dd>A JSON document that is augmented with support for Linked Data by providing an <code>@context</code> property with a defining URI [[!JSON-LD]].</dd>
-
   </dl>
   <p>
     <dfn data-lt="consume|consume a TD|consuming a TD">Consuming a Thing Description</dfn> means parsing the <a>Thing Description</a> and building a resource model with the use of <a>Protocol Bindings</a> that can be used in a script for accessing and controlling the Thing. A <a>Thing</a> when starts up, consumes its own <a>TD</a>, then it exposes its <a>WoT interface</a> as a server. Note that a script on a <a>Thing</a> may consume other <a>TD</a>s and expose a combined interface of the consumed <a>Things</a>.
   </p>
   <p>
+    A <dfn>WoT Runtime</dfn> or <dfn>WR</dfn> is defined as a script execution environment that manages the lifecycle of WoT application scripts, implements a script interpreter, an event loop, a security enforcement point for access management and uses an operating system API that provides access to local and remote resources. A <a>WR</a> should be isolated from other execution environments on memory address space, storage address space (file system), network namespace, etc.
```
h0ru5: optional suggestion (no impediment for merge, I can open an own PR) 
> "...and uses an operating system API that provides access to local and remote resources."

replaced by:
"...and uses lower-level APIs to provide access to local and remote resources."

Rationale: the runtime itself may be sandboxed

zolkis: Thanks, valid! Corrected.

rationale.md
```diff
@@ -2,6 +2,28 @@
...
+> The root WoT object should not be constructible. It represents the UA's magic ability to discover things, similar to how the Navigator object represents the UA's magic ability to do a bunch of stuff. A namespace might be a good replacement here, if it truly has no state.
+
+> Avoid constructor overloads. A true constructor should be something that directly copies the given essential data into internal fields. If there is a way to infer the essential data from some other data, then that should be a factory. So maybe ThingDescription is the essential data, and if we can infer that from a name or URL, then factory should be used (perhaps static factory, e.g. ExposedThing.fromName()).
+
+> The idea of using a builder pattern (first create an X, then call X.expose() on it to turn into another object, or X-with-UA-magic) is rather unidiomatic in JavaScript. The better way to represent something without UA magic is just a dictionary. Having the same class represent two very distinct things is not great.
+
+
+Resolutions:
+- The WoT API uses a namespace object `wot` in the browser, or provided by `require()` in standalone runtimes.
```
h0ru5: I understood the comment in this way: as an application developer, I want to be able to rely on the namespace to be magically there on global. If we want to give the option to explicitly require it, we can add the option that developers should be able to require (common.js) /  "import" (ES6) it.

If you agree, I would suggest the following formulation:
``- It is a runtime implementers' design choice whether the WoT API is accessible for application developers via a global namespace `wot` ( as usual in the e.g. browser), or by `require()` resp. `import` (as usual e.g. in standalone runtimes.)``

index.html
```diff
@@ -301,111 +301,78 @@
...
 
     dictionary ThingInit {
       DOMString name;
       USVString url;
-      object description;
+      Dictionary description;
     };
 
-    callback ThingDiscoveryCallback = void (ConsumedThing thing);
+    enum DiscoveryType { "any", "local", "directory", "broadcast" };
```
h0ru5: Bear with my limited knowledge of WebIDL: is this enum extensible? 

If not so, I'd suggest to have a list of well-known types, but the possibility to add any string here.
This is an important extension point IMO, as there are other discovery methods (e.g. reactive, such as geo, eddystone/physical web, ...)  we can pick up work from the discover TF to fill the well-known list, but I suggest we leave it extensible.

index.html
```diff
@@ -301,111 +301,78 @@
...
   </p>
   <p>
     The <dfn>WoT object</dfn> exposes only functions and has no internal state.
   </p>
-  <p class="issue">
-     Because of this, implementations MAY use a global object, or provide a constructor, or provide the API object, or by other means (e.g. by `require()`). Since the common denominator across various <a>Runtime</a> types would be to use a constructor, this would be recommended. As a consequence, the <a>ConsumedThing</a> and <a>ExposedThing</a> objects could also be constructed, complete with an asynchronous fetch functionality.
-  </p>
   <p class="note">
-    The WebIDL interfaces presented in this initial version of the document are based on the [[!WOT-PRACTICES]]. It is expected to be updated soon.
+     Browser implementations SHOULD use a namespace object such as `wot`, and [Node.js](https://nodejs.org/en/)-like runtimes MAY provide the API object through the [`require()`](https://nodejs.org/api/modules.html) mechanism.
```
h0ru5: See comment below. Rationale:
* can be global namespace for all kind of runtimes
* design decision  for implrmrmnter
* ES2015 ``import`` additionally to ``require()``

Suggestion:
`` Browser implementations SHOULD use a namespace object such as `wot`, and [Node.js](https://nodejs.org/en/)-like runtimes MAY alternatively provide the API object through the [`require()`] (https://nodejs.org/api/modules.html)  or [`import`](http://www.ecma-international.org/ecma-262/6.0/#sec-imports) mechanism.``

index.html
h0ru5: :+1:

index.html
```diff
@@ -417,42 +384,57 @@
...
     };
 
     ExposedThing implements ConsumedThing;
 
-    callback PropertyRetrieveHandler = void (ThingProperty property, USVString url);
-    callback PropertyUpdateHandler = void (ThingProperty property, USVString url);
-    callback ActionHandler = void (ThingAction action, USVString url);
-    callback ObserveHandler = void (DOMString eventName, USVString url);
+    callback PropertyHandler = void (ThingProperty property, USVString from);
+    callback ActionHandler = void (ThingAction action, USVString from);
```
h0ru5: correction: It should be ``Any`` instead of ``object`` for the param and the return.

index.html
zolkis: OK. Adding.

index.html
zolkis: No, enums are not extensible, if we wanted that would be just using a String.
However, this was intentional, as the supported discovery types should be standardized, since they require explicit and well defined support from the underlying platform and depend on well known standards.
There is no problem adding them to the spec, but they should be explicitly specified. Subsequent versions of the spec can extend the enum with new values.
Otherwise the API would be like `discover({ type: "app-specific-opaque-string"});` which is more like a free text search than scoping for supported discovery types.

index.html
```diff
@@ -417,42 +384,57 @@
     };
 
     callback ThingEventListener = void (ConsumedThing thing);
+
+    enum ConsumedThingEvents {
+        "propertychanged",
+        "propertyremoved",
+        "propertyadded",
+        "actionchanged"
```
zolkis: Did you mean `actionadded` and `actionremoved`, since `actionchanged` is there.
Yes, there might be value for clients to know there is different action or no action from now, or a new action in play.
I can add them.
We could also parametrize events, i.e. have only `propertychange` and `actionchange`, then the event data will carry the information whether was that adding, removing or updating.

index.html
```diff
@@ -417,42 +384,57 @@
...
 
       Promise&lt;void&gt; emitEvent(DOMString eventName, any payload);
 
-      // define request handlers
-      ExposedThing onPropertyRetrieve(PropertyRetrieveHandler handler);
-      ExposedThing onPropertyUpdate(PropertyHandler handler);
+      // define request handlers (one per request type, so no events here)
+      ExposedThing onRetrieveProperty(PropertyHandler handler);
+      ExposedThing onAddProperty(PropertyHandler handler);
+      ExposedThing onRemoveProperty(PropertyHandler handler);
```
zolkis: Indeed that is missing. I'd prefer the names `onUpdateProperty` and `PropertyUpdateHandler` since it matches better CRUDN terminology.

index.html
zolkis: I had slightly different understanding on how this works. Which use case do we implement here:
1. notify all observing clients if an action was called
2. notify all observing clients if an action was called, including which parameters was it called with and what was the return value?

Currently 1. is supported and you argue for supporting 2. It's fine if we have evidence this is a required use case.

rationale.md
zolkis: In the browser it shouldn't really be a choice. For runtimes, are you saying it can also be a namespace or global object? (plus obtainable by `require` or `import`)?

index.html
h0ru5: The risk for an application devloper is: "if I use a vendor-specific non-standard discovery method it will fail on a runtime that does not support this."

I see exactly this extensibility as a feature - have a well-defined of standardized discovery methods, but allow vendors to add own discovery methods that can later be standardized. That's how the browsers do & did evolve.

index.html
h0ru5: yes, sorry, typoed.
the parametrized event makes sense for me, but I'd think it could even lead to "interfacechanged" (or maybe "TDchanged") with the parameter like `{"method" : "add", "type" : "action", "name" : "bla"}`

Not sure which one is better: to have many simple events or fewer parameterized ones.

index.html
h0ru5: Yes, right.  I think we even had been naming it `onUpdateProperty` in current practices

index.html
h0ru5: Oh, Ok. I interpreted it as (3) setting the callback that gets called when the action is invoked.

h0ru5: from those cases i'd clearly opt for (1)

rationale.md
h0ru5: no, i mean the choice is: (a) like in the browser (b) require/import it

index.html
zolkis: In JavaScript they of course get mapped as String. The question is how do we describe in the algorithm of the `discover()` method. I prefer keeping (and extending now) the enum, eventually include an option for extended type but described in the algorithm. 
We should cover the known discovery types in this version. BLE and NFC could be covered by `"nearby"`. I think that should be enough since usually one method is used and the purpose of the filter is to rule out the others (e.g. broadcast type).

index.html
zolkis: I think it's simpler to write code for simple events. There is only 6 of them.

But I like the idea of having one event for `descriptionchange`. It is more expressive and creates less confusion (for instance with `propertychange` it is not clear whether the property value has changed, or the type, or the access). Of course we could have both `propertychange` (for TD) and `propertyupdate` (for value). But it would be more clear to have one `descriptionchange` event for TD changes for now.

zolkis: So we'd either have 
1) `propertychanged` (value), `descriptionchanged` (TD) and `actioninvoked`, 
2) or a flat list of 
`valuechanged` (value),
`propertychanged`, `propertyadded`, `propertyremoved`, `actionadded`, `actionchanged`, `actionremoved` (TD),
`actioninvoked`.

index.html
zolkis: Actually we don't need a separate `PropertyUpdateHandler` since in the request the `ThingProperty` parameter does have a requested new `value`, and the old value is what it is now, that is not changed. This is a request handler, not an event for notifications.

index.html
zolkis: Sorry, I messed it up. Disregard my previous comment. This is on the `ExposedThing`, and we are discussing the request handler for action related requests. The use case is that a request comes in from a client to invoke a local method. The `ThingAction` parameter contains all action related data: name, parameters, return value etc. The current WebIDL seems to be correct to me.

rationale.md
zolkis: I agree and this is what I meant in the text. So we need to make this more clear. What about this:

> The WoT API uses a namespace object `wot` in the browser. In non-browser runtimes, implementations may use either a namespace object `wot`, or an API object provided by the `require()` or `import()` mechanisms.

index.html
zolkis: I think 1. would be more efficient solution, because many TD changes could be expressed with one message, which also represents a TD change transaction, i.e. changes that needed to happen in the same time. Otherwise we cannot express this concept. So for the moment I will add the first event list.

zolkis:
Please check in the source the lines marked with `TBD`. 
There've been a lot of "consolidating" changes in the last commits, and these are the opens.

h0ru5:
As discussed among the editors, we will merge this PR  to create a state that can be the base for the F2F freeze resp. new detailed PRs if needed.

h0ru5:
Merged as discussed in the webconf on 02-05-2017

PR #24, opened 2017-05-06
added draft typescript definitions

h0ru5:
basically transpiling the webIDL into typescript

### DIFF
--- a/typescript-defs/package.json
+++ b/typescript-defs/package.json
@@ -1,0 +1,32 @@
+{
+  "name": "@types/wot",
+  "version": "0.0.1",
+  "description": "Typescript definitions for the WoT scripting API",
+  "main": "wot.d.ts",
+  "types": "wot.d.ts",
+  "typings": "wot.d.ts",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/w3c/wot-scripting-api.git"
+  },
+  "keywords": [
+    "web",
+    "WoT",
+    "IoT",
+    "scripting",
+    "typescript",
+    "types"
+  ],
+  "author": "Johannes Hund <johannes.hund@siemens.com>",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/w3c/wot-scripting-api/issues"
+  },
+  "homepage": "https://github.com/w3c/wot-scripting-api#readme",
+  "dependencies": {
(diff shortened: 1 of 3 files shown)


typescript-defs/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "keywords": [
+    "web",
+    "WoT",
+    "IoT",
+    "scripting",
+    "typescript",
+    "types"
+  ],
+  "author": "Johannes Hund <johannes.hund@siemens.com>",
+  "license": "MIT",
```
zolkis: Is MIT license compatible with W3C license?
https://github.com/w3c/wot-scripting-api/blob/master/LICENSE.md

typescript-defs/wot.d.ts
```diff
@@ -0,0 +1,182 @@
...
+    emitEvent(eventName : string, payload : any) : Promise<void>
+
+    // define request handlers (one per request type, so no events here)
+     onRetrieveProperty(handler : (PropertyRequest) => any) : ExposedThing
+     onUpdateProperty(handler : (PropertyRequest) => any) : ExposedThing
+     onInvokeAction(handler : (ActionRequest) => any) : ExposedThing
+     onObserve(handler : (ObserveRequest) => any) : ExposedThing
+}
+
+interface Request {
```
zolkis: Looks good, perhaps we need to update the WebIDL as well.

typescript-defs/wot.d.ts
```diff
@@ -0,0 +1,182 @@
...
+        name : string;
+        returnValue: any;
+    }
+}
+
+interface ThingDescriptionChangeEvent extends Event {
+    readonly data : {
+        name : string;
+        data: any; // : ThingPropertyInit | ThingActionInit | ThingEventInit 
+        type: TDChangeType;
```
zolkis: I think we use `any` when a property can be a simple type or object or `null`, so perhaps we could use the explicit alternatives with union types as marked in the comment, if it doesn't cause any issue. If you have reason to use `any`, we can keep it that way.

h0ru5: this was actually a leftover from editing it, the comment was meant to replace the any as soon as the types were defined

typescript-defs/package.json
h0ru5: AFAIK yes, but for the specs it would make sense to (dual-) license it under W3C license

zolkis:
+1 you can merge it :) Thanks!

PR #27, opened 2017-05-30
Fix WebIDL error.

zolkis:
`SemanticType[]` replaced with `sequence<SemanticType>`.


zolkis:
This is trivial, so merging.

PR #31, opened 2017-07-05
Add examples to the Scripting API spec

zolkis:


zolkis:
Looks like my branch is busy with commits. Earlier merges should have cleared them. Only the last commit  is valid.

zolkis:
Merging for the F2F reference - if there are issues, I will correct them later.

PR #37, opened 2017-08-07
ManagerThing&ReverseProxy

knimura:


ManagerThing.md
```diff
@@ -0,0 +1,142 @@
+# WoT ManagerThing
+
+The purpose of this document is to explain ManagerThing that can deal remote Thing lifecycle management by co-working with ScriptingAPI.
+
+### 1. ManagerThing
+
+ManagerThing is a capability in servient that provide management functionalities of servient and exposes management APIs to external clients.<br>
+
+The managerThing takes advantage of [Thing Description(TD)](https://w3c.github.io/wot-thing-description/), [ScriptingAPI](https://w3c.github.io/wot-scripting-api/), and [wot security-privacy](https://github.com/w3c/wot/tree/master/security-privacy).
```
zolkis: s/managerThing/ManagerThing

ManagerThing.md
```diff
@@ -0,0 +1,142 @@
+# WoT ManagerThing
+
+The purpose of this document is to explain ManagerThing that can deal remote Thing lifecycle management by co-working with ScriptingAPI.
+
```
zolkis: Perhaps some reformulation:
"This document presents a special Thing called `ManagerThing` that provides an interface to handle WoT scripts lifecycle management (install, update, run, stop, etc).

ManagerThing.md
```diff
@@ -0,0 +1,142 @@
...
+TD for ManagerThing declares fixed management commands as actions that a servient supports. <br>
+A Client that supports Scripting API(ConsumedThing) obtains the TD and knows what management capabilities are available. Then the Client issue commands to control the servient remotely.<br>
+
+### 2. Management commands
+ManagerThing provides the following managemet commands.<br>
+
+The following commands deals life cycle of script.
+* install: install a script and return a handle (e.g. URI, UUID).<br>
+* run: run a script by a handle.<br>
```
zolkis: add update?

ManagerThing.md
```diff
@@ -0,0 +1,142 @@
...
+* stop: stop a running script by a handle.<br>
+* uninstall: uninstall a script specified by a handle and discard the handle.<br>
+
+The script may include Scripting API commands.<br>
+
```
zolkis: What do you mean here? I think you could remove this line.

ManagerThing.md
```diff
@@ -0,0 +1,142 @@
...
+<img src=/images/ManagerThing_Fig1.png width=600 alt="Fig.1">
+
+The following diagram depicts how ManagerThing preapard to use.
+
+<img src=/images/ManagerThing1.png width=600 alt="Fig.2">
+
+- ManagerThing registers a TD for ManagerThing to a TD repository.
+- ConsumedThing in an external client acquires the TD by WoT. discover command.
+- ConsumedThing receives the TD and understands the what management capabilities are supported in the ManagerThing.
+
```
zolkis: I think we should mention how a ManagerThing gets into a runtime: they MUST be part of the SW image or SW package mechanism deployed to a Thing.

ManagerThing.md
```diff
@@ -0,0 +1,142 @@
...
+#### 5.2 Remote management of servient
+
+The external client manages a servient remotely using Scripting API(ConsumedThing).<br>
+- Servient has a ManagerThing and the ManagerThing is accessed from the client remotely.
+- ManagerThing inherited the mechanism from ExposedThing and expose WoT API for ManagerThing.
+
```
zolkis: I thought a ManagerThing was rather an instance of ConsumedThing... it's a fixed functionality on a remote device, and cannot be changed, just used.

zolkis:
Rendered version here:
https://github.com/knimura/wot-scripting-api/blob/master/ManagerThing.md

ManagerThing.md
zolkis: In other words, don't bother defining an ExposedThing for ManagerThing: they are not defined by a script, but get deployed with SW image. As a consequence, to simplify, revisit all sections where ExposedThing is associated with ManagerThing.

ManagerThing.md
```diff
@@ -0,0 +1,142 @@
...
+<img src=/images/ManagerThing_Fig2.png width=600 alt="Fig.3">
+
```
zolkis: IMHO you can simplify this picture: just show a ManagerThing, not the script that creates it (actually it's out of scope for the user how a ManagerThing is created, IMHO it just should be mentioned it's part of the SW image, and it is "there", and discoverable just as the other predefined Things on the device. You could have a separate section _in the end_, illustrating how to do a ManagerThing (by script, or natively part of the runtime).

ReverseProxy.md
```diff
@@ -0,0 +1,129 @@
+# WoT reverse proxy
+
```
zolkis: It would be better to call it a WoT directory (drawing parallel to [IETF resource directory](https://datatracker.ietf.org/doc/draft-ietf-core-resource-directory/)), or at least also mention  by this name.

zolkis:
Reverting, since review comments were not addressed.

PR #39, opened 2017-08-09
HTML fixes and first try to restructure WebIDL

danielpeintner:
currently WebIDL restructuring for Section 6 only

### DIFF
--- a/index.html
+++ b/index.html
@@ -92,7 +92,7 @@
 
   <section id="abstract">
     <p>
-      This specification describes a programming interface to [the Web of Things](http://w3c.github.io/wot/current-practices/wot-practices.html) (<dfn>WoT</dfn>), that allows scripts run on a <a>Thing</a> to discover and access other <a>Thing</a>s through a Client API, provide resources characterized by properties, actions and events through a Server API, and access locally attached hardware through a Physical API.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -92,8 +92,9 @@
 
   <section id="abstract">
     <p>
-      This specification describes a programming interface to [the Web of Things](http://w3c.github.io/wot/current-practices/wot-practices.html) (<dfn>WoT</dfn>), that allows scripts run on a <a>Thing</a> to discover and access other <a>Thing</a>s through a Client API, provide resources characterized by properties, actions and events through a Server API, and access locally attached hardware through a Physical API.
+      This specification describes a programming interface to [the Web of Things](http://w3c.github.io/wot/current-practices/wot-practices.html) (<code>WoT</code>), that allows scripts run on a <a>Thing</a> to discover and access other <a>Thing</a>s through a Client API, provide resources characterized by properties, actions and events through a Server API, and access locally attached hardware through a Physical API.
```
zolkis: I think using `<dfn>` instead of `<code>` has the advantage of linking possibility with `<a>`.

zolkis:
Looks good! Thanks!

index.html
danielpeintner: The issue in this specific case is that if we use here <dfn> we can't use it in the WebIDL later (duplicate dfn). That causes re-spec error and was the reason I removed it in the abstract.

PR #46, opened 2017-08-22
Add the Request interface, event definitions, expand descriptions.

zolkis:


index.html
```diff
@@ -414,24 +426,33 @@
...
+            <li>
+              "<dfn>other</dfn>" for a proprietary method defined by the solution.
+            </li>
           </ul>
         </p>
       </section>
     </section>
 
     <section> <h3>Examples</h3>
       <pre class="example" title="Discover Things via registry">
```
knimura: For the consistency to 5.3.2, I think EXAMPLE 1 can use the same term of “directory”, not “registry”.

+ <pre class="example" title="Discover Things via directory">

index.html
```diff
@@ -414,24 +426,33 @@
...
-        let discoveryType = { type: "registry", url: "http://wot.registry.org" };
+        let discoveryType = { method: "registry", url: "http://wot.registry.org" };
```
knimura: It can be:

+        let discoveryType = { method: "directory", url: "http://wot.directory.org" };

index.html
knimura: EXAMPLE 1: Discover Things via directory

index.html
```diff
@@ -736,27 +797,25 @@
           Emits an the event initialized with the event name specified by the <code>eventName</code> argument and data specified by the <code>payload</code> argument.
         </p>
         <p class="ednote">
-          This method in't actually needed to be exposed by the <a>ExposedThing</a> object since it is usually provided by the runtime.
+          This method isn't actually needed to be exposed by the <a>ExposedThing</a> object since it is usually provided by the runtime.
```
knimura: Regarding the EDITOR’S NOTE of 6.1.18, I think emitEvent() is necessary for servient app.
For example, a script count event from a device, then when reaching a threshold issue an emitEvent().
Such kind of app could use the command. I might not catch the statement correctly though, I think it can be removed.

zolkis: You're right that we need a method to notify observers, and that can be done as emitting an Event. 

But the note here just says the Scripting API does not need to explicitly expose this call, because emitting an event is always implemented in any JS runtime. So I am not sure why do we need to expose it this way. In most implementations, e.g. in Node.js we already have [emit()](https://nodejs.org/api/events.html#events_emitter_emit_eventname_args), now we will have an additional `emitEvent()` which does the same.

Do we differentiate between WoT events and runtime events - do we need to?

index.html
zolkis: I thing we can use either term, but let's stick with "directory" then.

danielpeintner:
A more general comment w.r.t. "1. Conformance". There the term language bindings is used.

> This specification can be used for implementing the WoT Scripting API in multiple language bindings. Currently ECMAScript and TypeScript definitions are described in this document. For specifying bindings in other languages, extensions of this document may be created later. 

Bindings is a bit confusing w.r.t. JavaScript/TypeScript (bindings in WoT has a different meaning--> protocol bindings)

What about changing it to "programming languages"

> This specification can be used for implementing the WoT Scripting API in multiple programming languages. Currently ECMAScript and TypeScript definitions are described in this document. For specifying other languages, extensions of this document may be created later. 

In Section "1. Conformance" bindings is used more often. Afterwards in the document bindings is just used for protocol bindings.

danielpeintner:
There is one more issue with the sections in the specification. 

"7.2 Security and Privacy" should be lifted on level to "8. Security and Privacy".

Similar with appendix sections that should read as
A.Changes
B. Open issues
C. Acknowledgements
D. References

zolkis:
> Bindings is a bit confusing w.r.t. JavaScript/TypeScript (bindings in WoT has a different meaning--> protocol bindings)

In W3C specs bindings usually refer to WebIDL being concerted to languages (JS, C++). 

We could use the full terms "language bindings" and "protocol bindings", in specific "WoT Protocol Bindings" which should be a defined term.

But we can also avoid the problem, as you have suggested. I will do that.

zolkis:
Links:
[htmldiff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fgithub.com%2Fw3c%2Fwot-scripting-api%2Fraw%2Fmaster%2Findex.html&doc2=https%3A%2F%2Fgithub.com%2Fzolkis%2Fwot-scripting-api%2Fraw%2Fmaster%2Findex.html)
[rendered version ](https://zolkis.github.io/wot-scripting-api/)

zolkis:
Up for proofreading again :).  @knimura  @danielpeintner

danielpeintner:
Thanks a lot for continuously integration our comments. Some more comments below


**A** 
https://zolkis.github.io/wot-scripting-api/#use-cases mentions
* Discover Things by filters defined on Thing Descriptions
* Discover Things by semantic filters.

With ThingFilter (being changed) having just method attribute this is not possible anymore.

**B**
https://zolkis.github.io/wot-scripting-api/#the-discover()-method
I am not sure why ThingFilter is optional. What should be discovered without any information?

**C**
https://zolkis.github.io/wot-scripting-api/#the-thingfilter-dictionary
We might want to add editorial note saying that ThingFilter may be extened with other attributes (as needed). 

**D**
https://zolkis.github.io/wot-scripting-api/#examples
Example 1, 3 and 4 make use of the afore mentioned (see **C**) extension with new attributes such as "url" and "description". Hence stating it makes sense.

**E**
https://zolkis.github.io/wot-scripting-api/#examples-0
in Example 5 it says 
console.log("Discovery finished because an error: " + error.message);
I think it should read 
console.log("Discovery finished because of an error: " + error.message);
or 
console.log("Discovery finished because an error '" + error.message + "' occurred.");

This error message is used in other places (Example 1, 2, 3, 4) also and might be fixed as well.

**F**
https://zolkis.github.io/wot-scripting-api/#examples-1
Example 7 and 8 are not syntax highlighted for whatever reason. It used to work for the currently published version (see https://w3c.github.io/wot-scripting-api/#examples-0)


**General**
URLs across the document need to be fixed/checked before FPWD but I think there is URL checker tool anyway (i.e. there are currently issues with "Repository" links in the very beginning of the document or in the Abstract).

zolkis:
A: No problem if we list use cases that are not supported in the current version yet. FPWD will elicit external comments on them and discussion may happen in issues to check which use cases the standard should address with what version roadmap.

B: ThingFilter is optional, with no args discovery will make the widest possible discovery allowed by the TD and Protocol Bindings. (the algorithm needs to say that indeed). Will check.

C: That is implicit with all Dictionary definitions, it doesn't need a note per se, but we can discuss on the next Editors Call  if it makes things more clear. 

D: Either the examples, or then the ThingFilter dictionary need to be fixed :). Which one should it be?

E: valid

F: will check

zolkis:
All the comments from above are fixed now, except the URLs that need checked.

zolkis:
NB: the repo reorg will be done after PR is merged, and it's time bound by tomorrow, so hopefully we can merge this tomorrow (Aug 30) about 8:00 UTC.
Chances for last comments are until then :). Thank you!

zolkis:
Merging now. Further issues will be fixed in separate PR.

PR #50, opened 2017-08-30
Remove obsolete files from the repo root

zolkis:


zolkis:
This is to correct the error in the previous PR where these files were not removed.

Issue #48, opened 2017-08-28
Clean up / reorganize the repo


zolkis:
I propose a cleanup of the scripting repository as follows:
- in the root we should have index.html, README.md
- other stuff should be moved into separate directories: primer, rationale, applications, applications/script-manager, applications/directory (formerly reverseproxy)
- under each directory there is index.html, README.md and the auxiliary resources (images, diagrams etc). This makes linking easy, and each contribution is contained.

- README.md and index.html is updated with new links.

PR #51, opened 2017-08-30
Repo cleanup. Consistency checks.

zolkis:
According to the Editors Call and the WG call on 30 Aug, 
- remove the HTML versions for primer, thing-directory and script-manager;
- remove TypeScript definitions for now.
- update ReSpec header for index.html for consistency with other specs.
- replace the Security section with a note and links to the IG security work and the WoT Architecture Security section.


zolkis:
Fixes issue #48

PR #53, opened 2017-09-06
update security section editor's note

mmccool:
Replace (broken, incomplete) security section with simple pointer to wot-security repo.  For some reason I can't merge and can't fetch the current state of the repo to resolve conflicts, however, so this PR will have to be merged manually.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1130,12 +1130,16 @@
     </section> <!-- ExposedThing Examples -->
   </section> <!-- The Thing Server API -->
 
-  <section> <h2 id="security">Security and Privacy</h2>
-    <p>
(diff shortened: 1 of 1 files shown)


mmccool:
Conflict seems to be due to terminology section immediately following security.   I'm not sure if it is supposed to be there and was just added, or was just deleted.   For whatever reason "fetch" is not retrieving it for me.   Zoltan, I'm sure you'll know which way this needs to be resolved...

mmccool:
I think I resolved the conflict and caught up my version with the head by preserving the (new) terminology section immediately after security.   Please confirm.

PR #60, opened 2017-09-18
Editorial changes

zolkis:
- Move FPWD files to a new `releases` directory
- Update README and CONTRIBUTING
- Add Johannes as former editor. Please note that ReSpec [doesn't yet support separate list for Former Editors](https://github.com/w3c/respec/issues/1097), therefore a note is used for specifying the current status. This can be changed later.


zolkis:
After this is merged, we can create a release/tag for the FPWD.

PR #65, opened 2017-10-09
ThingFilter uses non-existing/wrong enum

danielpeintner:
DiscoveryType does not exist --> now points to https://w3c.github.io/wot-scripting-api/#the-discoverymethod-enumeration


zolkis:
Thanks!

Issue #82, opened 2017-11-13
API for next PlugFest (Prague 2018)


mkovatsc:
```javascript
interface WoT {
    Observable<ConsumedThing> discover(optional ThingFilter filter);
    Promise<ThingDescription> fetchTD(USVString url);
    ConsumedThing             consume(ThingDescription td); // with fetch split out, we could make it synchrounous
    ExposedThing              expose(ThingTemplate init); // I think it was only symmetry to make it return a promise
};

typedef USVString ThingDescription;

// A full ThingTemplate aligned with ThingDescription requires pure JSON serialization of TD!
dictionary ThingTemplate {
    DOMString                   name;
    sequence<SemanticType>?     semanticTypes = []; // additional @types to "Thing" -- use case?
    sequence<SemanticMetadata>? metadata = []; // metadata fields in TD root (same level as 'name')
}

dictionary SemanticType {
    DOMString name;
    USVString context;
    DOMString prefix;
};

dictionary SemanticMetadata {
    SemanticType type;
    any          value;
};

interface ConsumedThing {
    readonly attribute DOMString name;
    readonly attribute ThingDescription td;
    Promise<any> readProperty(DOMString name); // aligned with 'writable'
    Promise<void> writeProperty(DOMString name, any value); // aligned with 'writable'
    Promise<any> invokeAction(DOMString name, any parameters);
    Observable onEvent(DOMString name);  // subscribe to events
    Observable onPropertyChange(DOMString name);  // subscribe to property change
    Observable onTDChange();  // subscribe to Thing Description change
};

interface ExposedThing: ConsumedThing {
    Promise<void> start(); // useful to make async or counter-productive?
    Promise<void> stop();
    Promise<void> register(USVString url); // convenience for Thing Directory
    Promise<void> unregister(USVString url); // convenience for Thing Directory
    Promise<void> emitEvent(DOMString eventName, any payload);
    
    void  addProperty(ThingPropertyInit property);  // throws on error
    void  removeProperty(DOMString name);  // throws on error
    void  addAction(ThingActionInit action);  // throws on error
    void  removeAction(DOMString name);  // throws on error
    void  addEvent(ThingEventInit event);  // throws on error
    void  removeEvent(DOMString name);  // throws on error
};

typedef DOMString ValueType;  // Linked Data JSON Schema

dictionary ThingPropertyInit {
    DOMString               name;
    ValueType               type; 
    any                     initValue;
    boolean?                writable = false;
[...]

Issue #78, opened 2017-10-26
API improvements


zolkis:
This issue is for discussing API improvements.

Let's start the discussion on observing functionality on the [ConsumedThing](https://w3c.github.io/wot-scripting-api/#the-consumedthing-interface) interface.

There we have
```javascript
Observable observe(DOMString name, RequestType requestType);

enum RequestType {  "property", "action", "event", "td" };
```
It allows subscribing and unsubscribing to notifications like:
- property change made by others
- action invoked by others
- any event supported by the target Thing
- changes to the Thing Description (TD) of the Thing.

Note that these could be also modeled with default events on [ExposedThing](https://w3c.github.io/wot-scripting-api/#the-exposedthing-interface), one being dedicated to 'descriptionchange', one for 'propertychange', one for 'actioninvocation' and then the signature would change to:

```javascript
Observable observe(DOMString eventName);
```
but the functionality would be implicit.

I wonder which would be more developer-friendly.

mkovatsc:
Where does the assumption come from that arbitrary Things do emit a notification when an Action was invoked? Is there any existing device already doing this?

zolkis:
It comes from the early WoT scripting API. We also discussed it in calls. @knimura, IIRC you needed it?

zolkis:
A simplified [`ConsumedThing`](https://w3c.github.io/wot-scripting-api/#the-consumedthing-interface) interface based on the FPWD version, but taking into account also #68:

```javascript
interface ConsumedThing {
    readonly attribute DOMString name;
    readonly attribute ThingDescription description;
    Promise<any>  invokeAction(DOMString name, any parameters);
    Promise<void> setProperty(DOMString name, any value);
    Promise<any>  getProperty(DOMString name);
    Observable    observe(DOMString name, optional ObserveOptions options);
};

callback ThingEventListener = void (ThingEvent event);

dictionary ObserveOptions {  // for conditional events
    callback filter = boolean (optional any value);
    // property-value pairs for constraints, defined by the TD
};

interface ThingEvent: Event {
    readonly attribute any data;
};
```

zolkis:
Alternatively, taking into account #64, `ConsumedThing` can provide an object representation of a Thing where Thing properties are mapped as object properties, Thing actions are mapped as functions, Thing Events are mapped as object events. In addition, Events can also be handled by `observe()` in order to support complex events such as conditional events, but this can be optional.

```javascript
interface ConsumedThing {
[...]

Issue #64, opened 2017-10-02
Simplified API for JavaScript and other programming languages


draggett:
## Properties

JavaScript supports getters and setters for object properties. This provides a means for accessing and assigning property values in a very direct way, e.g.

```
var p = room23.temperature;   // read the room's temperature property

room23.targetTemperature = 22;  // write to the targetTemperature property
```
The getters and setters would be automatically defined by the implementation of the functions exposed to applications for creating or consuming things as objects.

Some statically typed languages  (e.g, [Kotlin](https://kotlinlang.org/docs/reference/properties.html) and [Scala](http://www.dustinmartin.net/getters-and-setters-in-scala/)) provide syntactic support for getters and setters.

For other languages e.g. Java and Go lang, there are naming conventions for getter and setter methods, e.g.

```
double p = room23.temperature();   // read the room's temperature property

room23.setTargetTemperature(22);  // write to the targetTemperature property
```
You can easily find tutorials on the web e.g.

* http://www.codejava.net/coding/java-getter-and-setter-tutorial-from-basics-to-best-practices

Note that for go lang at least, the convention is to avoid using "get" as a prefix for getters, see:

* https://golang.org/doc/effective_go.html#Getters

JavaScript and Python have native support for getters and setters

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects#Defining_getters_and_setters
* https://www.python-course.eu/python3_properties.php

## Actions

A similar simplification can be used for thing actions, i.e. to define an action as a method on the object for a thing

Actions are invoked just like regular object methods, and return promises for the action response.

```
    thing42.someAction(args).then(response => console.log(response));
```

where "someAction" is the name of an action for thing42.

**Note** that JavaScript object properties and methods can't have the same names. The same should hold true for the properties and actions of things.

## Name spaces and observables

With the use of thing properties and actions as object properties and methods, we need a means to avoid name clashes with system properties and methods. One approach is to define a Web of things system object and to define methods for that object where a thing is passed as an argument. Another approach is to standardise a meta property for things that exposes an interface for such purposes as reading and updating thing metadata, and for registering observables for events.

The former approach is used below as a basis for applications to register handlers for events associated with an ECMAScript observable.

```
var subscription =  wot.observe(thing, observable, {
	next (value) { … },
	error (value) { … },
	complete () { … }
});
```

[...]

PR #86, opened 2018-01-29
Update the API according to discussion in #82 and #78

zolkis:
Don't merge yet, this is work in progress.
For the rendered version of the tip of the PR, use [this link](http://rawgit.com/zolkis/wot-scripting-api/master/index.html).

### DIFF
--- a/index.html
+++ b/index.html
@@ -123,13 +123,16 @@
       The Web of Things is made of entities (<a>Thing</a>s) that can describe their capabilities in a machine-interpretable format, the <a>Thing Description</a> (TD) and expose these capabilities through the <a>WoT Interface</a>. Support for scripting is optional for WoT devices.
     </p>
     <p>
-      By <a>consuming a TD</a>, a client <a>Thing</a> creates a runtime resource model that allows accessing the properties, <a>Actions</a> and <a>Events</a> exposed by the server <a>Thing</a>.
+      By <a>consuming a TD</a>, a client <a>Thing</a> creates a runtime resource model that allows accessing the <a>Properties</a>, <a>Actions</a> and <a>Events</a> exposed by the server <a>Thing</a>.
     </p>
     <p>
-      Exposing a <a>Thing</a> requires defining a <a>Thing Description</a> (TD) and instantiating a software stack needed to serve requests for accessing the exposed properties, <a>Actions</a> and <a>Events</a>. Defining a This specification describes how to expose and consume <a>Thing</a>s by a script.
+      Exposing a <a>Thing</a> requires defining a <a>Thing Description</a> (TD) and instantiating a software stack needed to serve requests for accessing the exposed <a>Properties</a>, <a>Actions</a> and <a>Events</a>. This specification describes how to expose and consume <a>Thing</a>s by a script.
     </p>
     <p class="note">
       Typically scripts are meant to be used on devices able to provide resources (with a <a>WoT interface</a>) for managing (installing, updating, running) scripts, such as bridges or gateways that expose and control simpler devices as WoT <a>Thing</a>s.
+    </p>
+    <p class="note">
+      This specification does not make assumptions on how the <a>WoT Runtime</a> handles and runs scripts, including single or multiple tenancy, script deployment and lifecycle management. The API already supports the generic mechanisms that make it possible to implement script management, for instance by exposing a manager <a>Thing</a> whose <a>Actions</a> (action handlers) implement script lifecycle management operations.
     </p>
     <p>
       For an introduction on how scripts could be used in <a>Web of Things</a>, check the [Primer](https://w3c.github.io/wot-scripting-api/primer) document. For some background on API design decisions check the [Rationale](https://w3c.github.io/wot-scripting-api/rationale) document.
@@ -138,7 +141,7 @@
 
   <section class="informative"> <h3>Use Cases</h3>
     <p>
-      The following scripting use cases are covered in this specification:
+      The following scripting use cases are supported in this specification:
     </p>
     <section><h4>Discovery</h4>
     <ul>
@@ -157,11 +160,14 @@
     <section><h4>Consuming a Thing</h4>
     <ul>
       <li>
-        Fetch and <a>consume a TD</a> of a <a>Thing</a>. The <a>TD</a> may be provided in a string serialized format, or specified by a URL. Consuming the <a>Thing</a> includes parsing the <a>TD</a> and attaching to the protocol bindings in order to access lower level functionality.
+        Fetch a <a>Thing Description</a> of a <a>Thing</a> given its URL.
+      </li>
+      <li>
+        <a>Consume a TD</a> of a <a>Thing</a>, including parsing the <a>TD</a> and generating the protocol bindings in order to access lower level functionality.
       </li>
       <li>On a consumed <a>Thing</a>,
         <ul>
-          <li>Get the value of a <a>Property</a> or set of properties.</li>
+          <li>Read the value of a <a>Property</a> or set of properties.</li>
           <li>Observe value changes to a <a>Property</a>.</li>
           <li>Set the value of a <a>Property</a> or a set of properties.</li>
           <li>Invoke an <a>Action</a>.</li>
@@ -174,13 +180,14 @@
     <section><h4>Exposing a Thing</h4>
     <ul>
       <li>
-        Exposing the <a>Thing</a> includes attaching to the protocol bindings in order to access lower level functionality.
+        Exposing the <a>Thing</a> includes generating the protocol bindings in order to access lower level functionality.
       </li>
       <li>
         Create a local <a>ExposedThing</a> to be exposed, based on a <a>Thing Description</a> provided in string serialized format, or out of a template or an existing <a>ConsumedThing</a> object.</li>
       <li>Add a <a>Property</a> definition to the <a>Thing</a>.</li>
       <li>Add an <a>Event</a> definition to the <a>Thing</a>.</li>
       <li>Add an <a>Action</a> definition to the <a>Thing</a>.</li>
+      <li>Attach semantic information to the <a>Thing</a></li>
       <li>Attach semantic information to an <a>Action</a></li>
       <li>Attach semantic information to a <a>Property</a></li>
       <li>Attach semantic information to an <a>Event</a></li>
@@ -188,7 +195,7 @@
         Emit an <a>Event</a>, i.e. notify all listeners subscribed to that <a>Event</a>.
       </li>
       <li>
-        Provide notifications for TD changes to clients subscribed to that.
+        Provide notifications for <a>TD</a> changes to clients subscribed to that.
       </li>
       <li>Mark/unmark the Thing to be discoverable.</li>
       <li>Mark/unmark the Thing to be consumable.</li>
@@ -201,31 +208,15 @@
           <li>
             to run an <a>Action</a>: take the parameters from the request, execute the defined action, and return the result;
           </li>
-          <li>to add a listener to an <a>Event</a>;</li>
-          <li>to remove an <a>Event</a> listener.</li>
-          <li>to fetch the <a>Thing Description</a>;</li>
         </ul>
       </li>
     </ul>
-    </section>
-    <section><h4>Future use cases</h4>
-    <p>
-      The following use cases are being considered out of scope for this version:
-    </p>
-    <ul>
-      <li>
-        Deploy, start, stop, delete scripts in local or remote runtimes in order to implement Thing lifecycle management.
-      </li>
-    </ul>
-    <p class="ednote">
-      Management use cases should be solved with exposing a special Thing that has an action for creating a <a>Thing</a> based on a <a>TD</a>.
-    </p>
     </section>
   </section>
 
   <section data-dfn-for="WoT">
     <h2>The <dfn>WoT</dfn> object</h2>
-    <p>The WoT object is the main API entry point and it is exposed by an implementation of the <a>WoT Runtime</a>. The <dfn>WoT object</dfn> has no internal state and provides methods for discovering, consuming and exposing a <a>Thing</a>.
+    <p>The WoT object is the API entry point and it is exposed by an implementation of the <a>WoT Runtime</a>. The <dfn>WoT object</dfn> does not expose properties, only methods for discovering, consuming and exposing a <a>Thing</a>.
     </p>
     <p class="note">
        Browser implementations SHOULD use a namespace object such as `wot`, and [Node.js](https://nodejs.org/en/)-like runtimes MAY provide the API object through the [`require()`](https://nodejs.org/api/modules.html) or [`import`](http://www.ecma-international.org/ecma-262/6.0/#sec-imports) mechanism.
@@ -448,18 +439,23 @@
 
     <section> <h3>Examples</h3>
       <pre class="example" title="Discover Things via directory">
-        let discoveryFilter = { method: "directory", url: "http://directory.wotservice.org" };
+        let discoveryFilter = {
+          method: "directory",
+          url: "http://directory.wotservice.org"
+        };
         let subscription = wot.discover(discoveryFilter).subscribe(
           thing => { console.log("Found Thing " + thing.url); },
           error => { console.log("Discovery finished because an error: " + error.message); },
           () => { console.log("Discovery finished successfully");}
         );
-        setTimeout(
-          () => { subscription.unsubscribe(); console.log("Discovery timeout"); },
+        setTimeout( () => {
+            subscription.unsubscribe();
+            console.log("Discovery timeout");
+          },
           5000);
       </pre>
       <p class="note">
-        Note that canceling a discovery (through unsubscribe) may not be successful in all cases, for instance when discovery is based on open ended broadcast requests. However, once `unsubscribe()` has been called,  implementations MUST suppress further event handling ( i.e. further discoveries and errors) on the Observable. Also, a discovery error may not mean the end of the discovery process. However, in order to respect Observable semantics (error always terminates processing), implementations MUST close or suppress further event handling on the Observable.
+        Note that canceling a discovery (through `unsubscribe()`) may not be successful in all cases, for instance when discovery is based on open ended broadcast requests. However, once `unsubscribe()` has been called,  implementations MUST suppress further event handling ( i.e. further discoveries and errors) on the Observable. Also, a discovery error may not mean the end of the discovery process. However, in order to respect Observable semantics (error always terminates processing), implementations MUST close or suppress further event handling on the Observable.
       </p>
       <pre class="example" title="Discover Things exposed by local hardware">
         let subscription = wot.discover({ method: "local" }).subscribe(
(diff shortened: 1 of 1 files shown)


mkovatsc:
This PR introduces many completely new concepts that conflict with our long-lived discussion in #82, e.g.:

* expose vs produce
* typedef ThingModel and reinterpretation of ThingTemplate
* ExposedThing without ConsumedThing interface (no control over the Thing)
* Cut down Interaction Inits
* dictionary ValueDefinition
* Additional set*Handler() functions

Is this to be changed as part of the work-in-progress, or are you proposing this now?

zolkis:
We have addressed these points during the call. Now the proposal is more or less clean. The reasons for change are the following.

1. expose vs produce
This is mainly a name change, coming from the fact that the returned object (ExposedThing) is not yet an exposed Thing, but an object that needs to  or might undergo changes such as adding properties/actions/events and request handlers. Please read the text, the current proposal allows creating an ExposedThing in 3 ways:
- based on a model (name + semantic annotations) that needs to be modified by adding properties/actions/events/handlers
- based on a TD, that needs to be added handlers
- based on a ConsumedThing (actually its TD, but implementations may do optimizations), that needs to be added handlers.

2. ThingModel reflects the possibilities from above, including ThingTemplate (that remained the same and in #82).

3. Actually a script has total control over an ExposedThing. It defines the thing, so cannot use it before it is defined. If ConsumedThing interface is needed, it needs to consume the ExposedThing's generated TD.

4.
> Cut down Interaction Inits

I am not sure what do you mean here. Perhaps 6. explains.

5. ValueDefinition is an attempt to align with the TD spec. Before it a full ThingDescription was used for specifying types - which was wrong and confusing.

6. The set*Handler() functions are needed for two reasons:
- it is an antipattern to specify function in init dictionary members
- they make explicit the decision that a TD does not contain scripts. Before, if you wanted to create and ExposedThing based on a template/TD, and wanted to add code (that was not present in the template/TD) you first need to remove e.g. a property/action, then re-add it with the handler mentioned in the init dict. This was just wrong. Now you can take a template/TD/ConsumedThing, replicate the TD-related structures and add the handlers on top. Also, it is clear at the top level that an ExposedThing must specify handlers, it is not any more hidden behind the members of a parameter of a function.

In fact, part of these changes are small, part of them are needed. The question is consistency, usability and alignment with the TD spec and other APIs. I believe this version is better (if not perfect) in that sense.

mkovatsc:
Mainly, we must not repeat the mishap of a broken (FP)WD, which was caused by ad hoc changes and introduction of unfinished, unconsolidated concepts. So why introduce these changes now and independently? The conclusion in the call was to write up what was discussed there.

1. expose vs produce -> fine by me
  Okay, small rename to be more precise, but also potentially more confusing in regard to established terminology.
2. ThingModel -> partly fine
  Ideally, ThingTemplate and ThingDescription would become the same; also only a subset of a TD is valid for exposing Things (e.g., not the binding information). Thus, `ThingTemplate` was introduced as intermediary solution. For now, we could allow for an object (ThingTemplate) or a string (ThingDescription), which is similar to an earlier version of the API.
  To put a ConsumedThing there seems to only fulfill a specific use case of optimizing proxy Thing creation and opens the path for the strange pattern of 3. This should not be introduced, definitely not for Prague because of the implications on the necessary implementation.
3. ExposedThing control -> not fine by me
  It is pretty ugly to juggle two objects, an ExposedThing and a ConsumedThing, to realize a fundamental use case: implementing the internals of a Thing. The `writeProperty()` function of an ExposedThing is different from the one of ConsumedThing: an ExposedThing is under the local control of the code and directly manipulates the state, while a ConsumedThing only sends request to modify a remote Thing under control of someone else. It was convenient to let ExposedThing implement the interface of ConsumedThing, but this example here tells me, it would be clearer to fully specify both interfaces independently. This was also raised in one of the calls with a similar motiviation.
4. Interaction Inits -> might be fine
  Weird that this comes up only now, as we are basically going back to the initial API (cf. `onInvokeAction()` etc.). This in itself might be a good sign, though.

mkovatsc:
@knimura and @danielpeintner please comment, so we can have a prepared call on Monday.

zolkis:
> 1. expose vs produce -> fine by me 

The remaining question is whether should we rather make it a constructor to ExposedThing instead of a factory method with disputed naming? Then we avoid the terminology discussion and have cleaner, more JS-like definition to create local ExposedThing objects. The only reason to keep the factory method would be if we wanted to create remote Things, but that is no longer a use case, as it can be done as a service.

> 2. ThingModel -> partly fine

I am also fine with removing ConsumedThing from the models. I was just thinking it could offer significant optimization opportunity when we want to re-expose a consumed thing, i.e. we don't need to fetch/consume a TD, then read the serialized form from the consumed thing and feed it to the ExposedThing factory that will parse it again. However, we could add this later, eventually.

> 3. ExposedThing control -> not fine by me

I am also fine with ExposedThing extending ConsumedThing, but we should be clear about the use cases.
The ExposedThing will define the properties, actions and events, then how to update/read a property, how to call an action, when to emit an event etc. So why would it want to use ConsumedThing methods to call an action, update/read a property and attach event listeners? It is implementing and exposing them, not using them as a client, not until finished exposing. There are two stages here, 1. exposing, 2. using (after consuming). I thought to make that distinction more clear. Perhaps we need to think what is the best way to expose the ConsumedThing interface on an ExposedThing, and when.

> 4. Interaction Inits -> might be fine

After playing with examples, I am pretty convinced now that this is the best way to go. I should add more example code to the spec.

mkovatsc:
1. Postpone discussing the constructor idea after Prague or prepare options for a F2F breakout.
2. Good, then let's limit to ThingTemplate dictionary or TD string. ConsumedThing.description can be passed and does not need a fetch.
3. It's not about extending ConsumedThing, but implementing the interface, as we need similar functions to control the local state. Consuming usually happens in a remote runtume. The implementation of the functions will be fundamentally different, yet I don't think we will need different names. At least postpone this discussion.
4. Let's bounce this of the other editors before the call.

knimura:
> 1. expose vs produce

No issues with the name change.

> 2. ThingModel – what is ThingModel

Not clear the definition of ThingModel. I have no problem with ThingTemplate, if the definition of ThingTemplate is the complete subset of TD but just excluding binding information from it.

> 3. “Actually a script has total control over an ExposedThing. It defines the thing, so cannot use it before it is defined. If ConsumedThing interface is needed, it needs to consume the ExposedThing's generated TD.”

I agree with ExposedThing extending ConsumedThing, because ExposedThing should have the same function as ConsumedThing. But maybe difficult to understand the notion of "inheriting client API of local thing is part of ExposedThing(ServerAPI)", so showing the use case would be needed.

> 5. ValueDefinition is an attempt to align with the TD spec. Before this a full ThingDescription was used for specifying types - which was wrong and confusing.”

Totally agree.


> 6.	“The set*Handler() functions are needed for two reasons:
> If you wanted to create and ExposedThing based on a template/TD, and wanted to add code you first needed to remove e.g. a property/action, then re-add it with the handler mentioned in the init dict. 
	
Agree. If ThingTemplate is subset of TD and ExposedThing is created from ThingTemplate, both removeProperty and addProperty must be called to setting handler. If we feel that going back to the initial API, introducing UpdateProperty may be one of the solution to avoid onRetrieveProperty or so.

zolkis:
> Not clear the definition of ThingModel. 

A Thing model (the term was taken from #64) is either a ThingTemplate (dictionary) or a ThingDescription (string).

I have changed back to `ExposedThing implements ConsumedThing`.

zolkis:
As discussed on the call, on request by @knimura , added a short informal section for Observables, and on  request by @danielpeintner , removed optionality from some of the arguments and dictionary members.

zolkis:
As discussed in the call, added explanations on how handlers are working/implemented on ExposedThing. @mjkoster please take a look.

danielpeintner:
One comment/question: I noticed that PropertyWriteHandler  gets only the new value. We used to have "oldValue" and "newValue". Any rational for this change?

zolkis:
WiP is removed, proof-reading can start.

mkovatsc:
> One comment/question: I noticed that PropertyWriteHandler gets only the new value. We used to have "oldValue" and "newValue". Any rational for this change?

This is a quite strong concept change. As I mentioned, we had a reactive programming pattern before, which has at least two benefits:

* Script developers do not need to care about proper and safe state management. Without setting a PropertyHandlers, we have a basic resource behavior (set state with PUT, read state with GET).
* Better performance, as requests can be answered directly from the runtime space, no need to hand over to the scripting space (which was a bit problematic with our previous Java + Nashorn implementation and will be problematic for embedded runtimes).

Such a change must be explicitly discussed and is not fit for Prague.

zolkis:
There is an editor's Note in the spec stating:

> Note that this function is invoked by implementations before the property is updated, so the client code in this function can invoke the readProperty() method to find out the old value of the property, if needed. Therefore the old value is not provided to this method.

So your use case seems to be solved.

mkovatsc:
We are discussing server side -- the client has nothing to do with these handlers.

zolkis:
Yes, server side. Client there means client to the server implementation (servient), i.e. the "guest code" in the callback.

mkovatsc:
Okay, I see you already removed "client" from that note ;)

Have you checked that the handler code has always the right reference to the ExposedThing object to call `readProperty(propertyName)`? To me it is not obvious without testing, also function vs lambda...

zolkis:
Of course the handler code has the right reference to the ExposedThing object.
```javascript
thing.setPropertyWriteHandler(value => {
  thing.readProperty(propertyName).then(oldValue => {
    // put handler code here
   if (oldValue < 5) {
      // update to value 
      gpio.port.write(0x02, value);
   } 
  });
}, propertyName);
```

index.html
```diff
@@ -262,29 +240,60 @@ <h2>The <dfn>WoT</dfn> object</h2>
...
           The <a>ThingFilter</a> dictionary that represents the constraints for discovering <a>Thing</a>s as key-value pairs.
         </p>
         <pre class="idl">
           dictionary ThingFilter {
-            (DiscoveryMethod or DOMString) method = "any";
+            DiscoveryMethod method = "any";
             USVString url;
-            Dictionary description;
+            USVString query;
+            sequence&lt;Dictionary&gt; constraints;
```
danielpeintner: I wonder why we need  a `sequence<Dictionary>`.
According to my understanding `Dictionary constraints` seems sufficient.

danielpeintner: The AND and OR relationship for constraints seems to be very powerful but also complex at first. Moreover, the link to "the semantic metadata" in its description is very fuzzy. As I understand it, the key of the key-value pair is a reference to `SemanticType`. If so we might want to add an example.

zolkis: Example 4 is such an example: discover nearby things by Bluetooth OR NFC.

index.html
```diff
@@ -293,80 +302,160 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
...
+        Accepts a <code>model</code> argument of type <code><a>ThingModel</a></code> and returns an <a>ExposedThing</a> object, locally created based on the provided initialization parameters. An <a>ExposedThing</a> can be created in the following ways:
+        <ul>
+          <li>
+            from an initial model (name and semantic annotations), then adding properties, actions, events and request handlers;
+          </li>
+          <li>
+            from a <a>Thing Description</a>, then adding request handlers;
+          </li>
+          <li>
+            from an existing <a>ConsumedThing</a> object, based on its <a>Thing Description</a>, then adding request handlers.
```
danielpeintner: The third option is a special case of option 2. Hence I propose to remove option 3 and change option 2 to 

from a Thing Description (possibly also retrieved from a ConsumedThing), then adding request handlers;

zolkis: That should be fine. I just wanted to highlight it is possible to expose a ConsumedThing.

zolkis: Done (not committed yet).

index.html
```diff
@@ -817,32 +723,26 @@ <h4>The <dfn>SemanticType</dfn> dictionary</h4>
...
         <pre class="idl">
-          dictionary ThingActionInit {
-            DOMString name;
-            ThingDescription inputDataDescription;
-            ThingDescription outputDataDescription;
-            sequence&lt;SemanticType&gt; semanticTypes;
-            Function action;
+          dictionary ThingActionInit: SemanticAnnotations {
+            required DOMString name;
+            ValueType inputDataDescription;
```
danielpeintner: Align with TD draft, name it `inputData` and `outputData`?

index.html
```diff
@@ -853,105 +753,80 @@ <h4>The <dfn>ThingActionInit</dfn> dictionary</h4>
       <section data-dfn-for="ThingEventInit">
         <h4>The <dfn>ThingEventInit</dfn> dictionary</h4>
         <pre class="idl">
-          dictionary ThingEventInit {
-            DOMString name;
-            sequence&lt;SemanticType&gt; semanticTypes;
-            ThingDescription dataDescription;
+          dictionary ThingEventInit: SemanticAnnotations {
+            required DOMString name;
+            ValueType dataDescription;
```
danielpeintner: Align with TD draft, name it `outputData`?

index.html
```diff
@@ -962,31 +837,46 @@ <h2>Examples</h2>
       <pre class="example highlight" title="Create a new blank exposed Thing">
         WoT.createExposedThing(thingDescription)
```
danielpeintner: Old API example. Should be `WoT.produce` without Promise and `then` etc

Also in addAction there is still old handler code.

index.html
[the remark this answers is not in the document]
zolkis: Done.

index.html
```diff
@@ -293,80 +302,160 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
...
 
     <section> <h3>Examples</h3>
       <pre class="example" title="Discover Things via directory">
-        let discoveryFilter = { method: "directory", url: "http://directory.wotservice.org" };
+        let discoveryFilter = {
+          method: "directory",
+          url: "http://directory.wotservice.org"
+        };
         let subscription = wot.discover(discoveryFilter).subscribe(
           thing => { console.log("Found Thing " + thing.url); },
```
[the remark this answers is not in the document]
zolkis: Done.

index.html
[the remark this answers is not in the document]
zolkis: See Example 4. ("protocol" is made up for the example)

index.html
```diff
@@ -375,15 +464,28 @@ <h4>The <dfn>ThingInit</dfn> dictionary</h4>>
...
+        let subscription = wot.discover({ method: "local" }).subscribe({
+          next: thing => { console.log("Found local Thing " + thing.url); },
+          error: err => { console.log("Discovery error: " + err.message); },
+          complete: () => { console.log("Discovery finished successfully");}
+        });
+      </pre>
+      <pre class="example" title="Discover Things exposed nearby, via Bluetooth or NFC">
+        let subscription = wot.discover({
+          method: "nearby",
+          constraints: [{ protocol: "BLE-4.2" }, { protocol: "NFC"}]
```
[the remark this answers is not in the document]
zolkis: These examples are made up, at the moment we don't have a standard attribute "protocol", but it could be done if it made sense in the Thing context.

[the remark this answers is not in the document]
zolkis: Yes,  the constraints MAY be of type `SemanticAnnotation` (type + value), but in general they don't need to be. They are (at the moment) key -value pairs (dictionaries) that can be mapped to `SemanticAnnotation` like 
```javascript
let constraint = { 
    semanticTypes: [ { name: "protocol", context: "..." } ], 
    metadata: [ { type: "protocol", value: "BLE-4.2" } ]
};
```
which is very developer-unfriendly way to tell a constraint. Instead, the implementation of the Scripting API can translate the constraint dictionaries to SemanticAnnotation.

index.html
```diff
@@ -395,30 +497,26 @@ <h4>The <dfn>ThingInit</dfn> dictionary</h4>>
...
     <p>
-      The <a>ConsumedThing</a> interface is a client API for sending requests to servers in order to retrieve or update properties, invoke <a>Actions</a>, and observe properties, <a>Actions</a> and <a>Events</a>.
+      The <a>ConsumedThing</a> interface is a client API for sending requests to servers in order to retrieve or update <a>Properties</a>, invoke <a>Actions</a>, and observe <a>Properties</a>, <a>Actions</a> and <a>Events</a>.
     </p>
     <pre class="idl">
       interface ConsumedThing {
         readonly attribute DOMString name;
-        readonly attribute USVString url;
-        readonly attribute ThingDescription description;
+        readonly attribute ThingDescription td;
```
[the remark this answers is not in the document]
zolkis: A property has a getter function that can be the `getTd()` - in my view they are equivalent, aren't they?
Of course we could even have a `Promise<ThingDescription> getTD()`, but do we need it?

index.html
```diff
@@ -634,111 +599,76 @@ <h2>Examples</h2>
...
+        void removeAction(DOMString name);
+        void addEvent(ThingEventInit event);
+        void removeEvent(DOMString name);
+        void setActionHandler(ActionHandler action, optional DOMString actionName);
+        void setPropertyReadHandler(PropertyReadHandler readHandler, optional DOMString propertyName);
+        void setPropertyWriteHandler(PropertyWriteHandler writeHandler, optional DOMString propertyName);
       };
-      ExposedThing implements ConsumedThing;
+      callback ActionHandler = Promise&lt;any&gt; (any parameters);
+      callback PropertyReadHandler = Promise&lt;any&gt; (DOMString propertyName);
```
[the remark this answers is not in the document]
zolkis: Right, we discussed this and I haven't removed them. Done now.

index.html
```diff
@@ -749,20 +679,22 @@ <h3>The <dfn>RequestHandler</dfn> callback</h3>
       <section data-dfn-for="ThingPropertyInit">
         <h4>The <dfn>ThingPropertyInit</dfn> dictionary</h4>
         <pre class="idl">
-          dictionary ThingPropertyInit {
-            DOMString name;
+          dictionary ThingPropertyInit: SemanticAnnotations {
+            required DOMString name;
+            required ValueType type;
+            any value;
             boolean configurable = false;
```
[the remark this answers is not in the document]
zolkis: Done.

index.html
zolkis: `inputData` would imply this is data, but it is a description of input data.

index.html
zolkis: `outputData` is a bad name, usually events have "data" in JS. This API is for developers.

index.html
zolkis: Thanks for the catch. Done.

zolkis:
There is still a lot of work needed on SemanticAnnotations and DataSchema, but for the moment this is good to be merged.

zolkis:
I get a ReSpec warning for undefined Observable, but it's defined. Someone please take another look and try to fix.

danielpeintner:
Merging is fine by me also!

Note: the ReSpec issue you are seeing relates to the observable attribute in "5.7.1 The ThingPropertyInit dictionary". The name collision seems to cause that issue. Renaming the attribute made the ReSpec issue disappear.

Let's try to look into after merging...

zolkis:
This is merged now. Thank you all for the comments and reviews.

PR #98, opened 2018-03-03
Change ThingProperty defaults

mkovatsc:
Default false is more sensible for writeable and even more for observable

### DIFF
--- a/index.html
+++ b/index.html
@@ -699,8 +699,8 @@
             required DOMString name;
             required DataSchema schema;
             any value;
-            boolean writable = true;
-            boolean observable = true;
+            boolean writable = false;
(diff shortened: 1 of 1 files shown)


zolkis:
Why does it make more sense?
Please update the text as well, and check if the examples need updating, too.

mkovatsc:
One reason is right in the [WebIDL spec](https://www.w3.org/TR/WebIDL-1/#dfn-optional-argument-default-value).
And then writeable and, in particular, observeable are additional features for Properties. Why would the default be enabled?

zolkis:
Reviewed, no further comments until now, merging.

PR #99, opened 2018-03-03
Use singular for field names (LD convention)

mkovatsc:
The field `semanticTypes` of SemanticAnnotations should be `semanticType`.


zolkis:
Update the text as well.

zolkis:
Trivial change, reviewed, merging.

PR #101, opened 2018-03-04
Fix examples

mkovatsc:
Changed examples to new API.
Tested with node-wot, except `setPropertyReadHandler()`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -847,90 +847,77 @@
       </p>
 
       <pre class="example highlight" title="Create a new blank exposed Thing">
-        let thing = WoT.produce(thingDescription);
         try {
-          // manually add properties, actions, and events
+          var thing = WoT.produce({ name: "tempSensor" });
+          // manually add Interactions
           thing.addProperty({
             name: "temperature",
-            value: "0",
-            schema: '{ \"type\": \"number\" }',
-            writable: false
+            value: 0.0,
+            schema: '{ "type": "number" }'
             // use default values for the rest
-          });
-          thing.addEvent({
-            name: "onchange",
-            schema: '{ \"type\": \"number\" }'
-          });
-          thing.addAction({
+          }).addProperty({
+            name: "max",
+            value: 0.0,
+            schema: '{ "type": "number" }'
+            // use default values for the rest
+          }).addAction({
             name: "reset",
-            action: () => { this.temperature = 0; }
+            // no input, no output
+          }).addEvent({
+            name: "onchange",
+            schema: '{ "type": "number" }'
           });
           // add server functionality
-          thing.setPropertyReadHandler(propertyName => {
-            console.log("Handling read request for " + propertyName);
-            return new Promise((resolve, reject) => {
-                if (this.hasOwnProperty(propertyName))
-                  resolve(this[propertyName]);
-                else
-                  reject('Property not found');
-              },
-              e => {
-                console.log("Error");
-              });
-          });
-        } catch (err) {
-           console.log("Thing creation error.");
-        }
-        thing.start()
-        .then (() => {
-          thing.register();
-        });
-      </pre>
+          thing.setActionHandler( () => {
+            console.log("Resetting maximum");
+            thing.writeProperty("max", 0.0);
+          }, "reset");
 
-      <pre class="example highlight" title="Create a new exposed Thing from a TD URI">
-        let thingDescription = '{
-          "name": "mySensor",
-          "url": "http://myservice.org/mySensor/description"
-        }';
-        WoT.fetch("http://myservice.org/mySensor/description")
-        .then(td => {
-          let e_thing = WoT.produce(td);
-          // properties, actions and events have been added based on the TD
-          // now add the requests handlers
-          try {
-            e_thing.setPropertyReadHandler(propertyName => {
-              console.log("Handling read request for " + propertyName);
-              return new Promise((resolve, reject) => {
-              if (this.hasOwnProperty(propertyName))
-                resolve(this[propertyName]);
-              else reject('Property not found');
-            });
-            e_thing.start();
-          } catch(error) {
-            console.log("Error");
-          }
+          thing.start();
+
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -847,90 +847,77 @@ <h2>Examples</h2>
...
+            let old = await thing.readProperty("max");
+            if (old < mock) {
+              thing.writeProperty("max", mock);
+              thing.emitEvent("onchange");
+            }
+          }, 1000);
+
         } catch (err) {
-           console.log("Thing creation error.");
+           console.log("Script error: " + err);
```
zolkis: Needs to be a specific error. In this case the `try` block is about creating the Thing, so the old version was more appropriate.

index.html
```diff
@@ -847,90 +847,77 @@ <h2>Examples</h2>
...
         }
-        thing.start()
-        .then (() => {
-          thing.register();
-        });
       </pre>
```
zolkis: You removed the `start()` and `register()` part. It is important the reader understands that producing a Thing does not mean the Thing is running or is discoverable.

index.html
```diff
@@ -847,90 +847,77 @@ <h2>Examples</h2>
...
+        let thingDescription = '{ "@context": [ "https://w3c.github.io/wot/w3c-wot-td-context.jsonld", "https://w3c.github.io/wot/w3c-wot-common-context.jsonld" ], "@type": [ "Thing", "Sensor" ], "name": "mySensor", "geo:location": "testspace", "interaction": [ { "@type": [ "Property", "Temperature" ], "name": "prop1", "schema": { "type": "number" }, "saref:TemperatureUnit": "degree_Celsius" } ] }';
+        try {
+          // note that produce() fails if thingDescription contains error
+          let thing = WoT.produce(thingDescription);
+          // Interactions were added from TD
+          // add server functionality
+          thing.setPropertyReadHandler( (propertyName) => {
+            console.log("Handling read request for " + propertyName);
+            return new Promise((resolve, reject) => {
+              resolve(Math.random(100));
```
zolkis: You need the test of `hasOwnProperty`.

zolkis: The original example was better IMHO.

index.html
mkovatsc: Note that runtimes have a log, too. In that context, using an actual implementation, "script error" is actually more precise and helpful.

index.html
mkovatsc: Yes forgot. (Mainly because there is no real use case behind this.)

index.html
mkovatsc: Where does `hasOwnProperty` come from? So far it is not defined.

Why go the way of a single handler with your own multiplexer in the first place, when you rely on the Thing again to manage state and presence?

When the handler is given for a concrete Interaction, no check is needed (this case).

When the handler is for all Interactions, then the handler must be able to check is it holds the state. Experience on this is available from LWM2M implementation.

zolkis: From ECMAScript.

zolkis: Originally we had the names provided, but we discussed in the scripting call with Daniel and turned out that in a real script context we don't need them. Could you provide examples where you can't figure out the context without the names provided?

index.html
zolkis: Then the scripting implementation will provide that context, so that in the runtime log it appears as coming from a script. In scripts the error context is the scripting API, and not the full WoT runtime.
Saying "script error" from a script is equivalent to "error".

index.html
mkovatsc: Actually, there is a `thing.start()` further up in the example (before setting the interval to act upon the Thing).

Is `register()` also mandatory to call?

index.html
mkovatsc: If think you are discussing the generic handlers here, while the example shows a specific handler for `prop1`. As discussed in the call, I would not even model the `propertyName` argument for this specific handler.

I guess what you want is an example for a generic handler that shows the multiplexing based on the passed name?

zolkis: No, the original example was about a generic handler that performs the same things for all property reads. Given the property name, it fetches and then returns the property value. Hence you don't need to write (or bind) the same function with each and every property supported (suppose there are 20 properties). It just saves space. But this was not meant for providing a multiplexed implementation for a given interaction type. That was not supported.

zolkis: I suggest we separate these examples, i.e.have one as generic handler without the need to refer to a certain property (the original example), the other for in-handler dispatching of property handling, and another for specific handler that is not using the property name (since it was specifically bound to a property).

mkovatsc: So at least: yes, you wanted an example for a generic handler.

If there are 20 Properties, I don't see the point of returning the same result for all of them. To individualize, you need the multiplexing I am talking about (based on the propertyName passed to the handler).

I would add a generic handler for two different Actions. Having a generic handler for a single Property does not drive the point.

zolkis: It's not returning the same result! Read the code you have removed.
Internally, there are slots for all properties. The handler gets the name. So the implementation can fetch the value of the requested one. That is not multiplexing. It's parametrization. Multiplexing is another example we did not have so far.

index.html
```diff
@@ -847,90 +847,77 @@ <h2>Examples</h2>
...
-            // ...
-          }'
-        };
-        let e_thing = WoT.produce(thingDescription);
-        // properties, actions and events are added based on the TD
-        // ...
+      <pre class="example highlight" title="Create a new exposed Thing from a TD URI">
+        // fetch an external TD, e.g., to set up a proxy for that Thing
+        WoT.fetch("http://myservice.org/mySensor/description").then(td => {
+          // WoT.produce() ignores instance-specific metadata (security, form)
```
zolkis: This behavior is not specified yet, so should not be (yet) mentioned in an example.

zolkis:
Merging this now and fix the issues in a later PR.

index.html
mkovatsc: That was an existing example that I only fixed, not add.

Issue #102, opened 2018-03-04
Callbacks need to take Interaction name


mkovatsc:
For the feature to set a single handler for all Interactions of a time, the Interaction name has to be passed at well.

Should be reflected in the callback listing:
```javascript
callback PropertyReadHandler = Promise<any> (string propertyName);
callback PropertyWriteHandler = Promise<void> (any value, string propertyName);
callback ActionHandler = Promise<any> (any parameters, string actionName);
```

The "The ActionHandler callback" section text also needs update. The other two sections look good.

zolkis:
When a name is provided, the handler is associated with that name.

When no name is provided, the handler is associated with all names of that interaction type (i.e. it is a generic handler).

This simplifies coding a great deal. You can set a generic handler, and then only override the ones you want.

zolkis:
Actually we have discussed on a past scripting call whether name should be passed to callbacks, and we agreed it's not needed. The `invokeAction()` is passed the name, not the callback.

mkovatsc:
My point is that at the moment a single handler cannot know for which Interaction it was called. The current status is:

```javascript
callback ActionHandler = Promise<any> (any parameters);
callback PropertyReadHandler = Promise<any> ();
callback PropertyWriteHandler = Promise<void> (any value);
```

zolkis:
To summarize:
- when a specific handler is provided, the callbacks don't need the interaction name
- when a generic handler is provided, the original idea was that it should be generic, i.e. not bound to interaction name, or it's the common part that is done for all interactions for that type.
- however, this issue is asking for supporting generic handlers which internally dispatch handling to sub-handlers based on the interaction names. In that case, we do need to add interaction names to handler callbacks.

As a result:
1. When the `set...Handler()` are called without the interaction names (i.e. generic handlers for that interaction type), the callbacks will always have to receive the interaction name from implementations.
2. When the `set...Handler()` are called with the interaction names bound to the provided function, the callbacks can be provided as functions that do not receive the interaction name.

Now the question is which parameter should the interaction name be: first or last (in order to harmonize with the `set...Handler()` functions. @mkovatsc proposal was to put them as optional second argument to callbacks.

[...]

PR #103, opened 2018-03-07
Fix handler callbacks WebIDL, update text and examples.

zolkis:


danielpeintner:
Do not understand what `this.hasOwnProperty(propertyName)` is for in 

```
  thing.setPropertyReadHandler(propertyName => {
    console.log("Handling read request for " + propertyName);
    return new Promise((resolve, reject) => {
        if (this.hasOwnProperty(propertyName))
          // application can dispatch fetching property values
          // if (propertyName === 'prop1') { ... }
          // but in general, just read the property value
          resolve(this[propertyName]);
        else
          reject('Property not found');
      },
      e => {
        console.log("Error");
      });
  });
```

I think we cannot put dispatcher code there. According do my understanding the snippet above should be simply

```
  thing.setPropertyReadHandler(propertyName => {
    console.log("Handling read request for " + propertyName);
    return new Promise((resolve, reject) => {
          // application can dispatch fetching property values
          // if (propertyName === 'prop1') { ... }
          // but in general, just read the property value
          resolve(...);
    });
  });
```

zolkis:
@danielpeintner @knimura @mkovatsc  could any of you review please?

index.html
```diff
@@ -803,37 +803,37 @@ <h3>The <dfn>PropertyWriteHandler</dfn> callback</h3>
...
       </p>
       <p>
-        If provided, this callback function will implement reading a <a>Property</a> and SHOULD be called by implementations when a request for reading a <a>Property</a> is received from the underlying platform.
+         The <code>readHandler</code> callback function will implement reading a <a>Property</a> and SHOULD be called by implementations when a request for reading a <a>Property</a> is received from the underlying platform.
       </p>
       <p>
-        There SHOULD be at most one handler for any given <a>Property</a>. If no handler is initialized for any given <a>Property</a>, implementations SHOULD implement default property read without calling this hook.
+        There SHOULD be at most one handler for any given <a>Property</a> and newly added handlers replace the old handlers. If no handler is initialized for any given <a>Property</a>, implementations SHOULD implement a default property read handler.
```
danielpeintner: Minor comment: in node-wot we deal with any property without a readHandler by simply returning the value as is (last set value). Hence I am a bit unsure about the 2nd sentence.  Does it make sense to simplify the 2nd sentence above to something like this?

"If no handler is provided for a given <a>Property</a>, implementations are required to report the property value as is (e.g., set by  writeProperty() or ThingProperty.value)"

index.html
```diff
@@ -803,37 +803,37 @@ <h3>The <dfn>PropertyWriteHandler</dfn> callback</h3>
...
+        Takes <code>name</code> as string argument and <code>writeHandler</code> as argument of type <a>PropertyWriteHandler</a>. Sets the handler function for writing the specified <a>Property</a> matched by <code>name</code>. Throws on error. Returns a reference to the same object for supporting chaining.
       </p>
       <p>
-        There SHOULD be at most one write handler for any given <a>Property</a>. If no write handler is initialized for any given <a>Property</a>, implementations SHOULD implement default property update and notifying observers.
+        There SHOULD be at most one write handler for any given <a>Property</a> and newly added handlers replace the old handlers. If no write handler is initialized for any given <a>Property</a>, implementations SHOULD implement default property update and notifying observers on change.
```
danielpeintner: Similar to the previous comment the 2nd sentence might look like

"If no write handler is initialized for a given <a>Property</a>, implementations SHOULD set the property value with the provided new data  and notify observers."

Maybe "default property update" means the same to you. IF so fine also.

zolkis: My thinking was that the handler actually implements writing the property in a way known only to it (e.g. using local call such as GPIO or network etc). Or it may only want to implement some side-effects (hooks) to setting the property value. Of course the WoT runtime can always update the "proxy" property value it maintains (i.e. in your case you only have the "proxy" property value, basically a memory location that is not bound to HW).

That is why I used the term "default property update", to leave open what that actually means (until we define the algorithms), since it may be more than just updating a local memory location.

danielpeintner: Agree. Makes sense. Let's leave it as is for now.

zolkis:
Fixes #102.

PR #105, opened 2018-03-25
Update w3c.json

steventam312:


dontcallmedom:
this breaks the expected format of w3c.json, closing

Issue #112, opened 2018-04-22
API must allow to specify concrete media type


mkovatsc:
Usually the WoT Runtime decides which `mediaType`s to offer in the `forms` section, as it depends on the representation formats implemented (serializers/parsers). This is good for abstract data with a schema that can represented in multiple formats such as JSON or CBOR.

Some applications require to offer specific representation formats, in particular when it is not the usual abstract data with a schema such as binary data. Examples are images or certificates in DER format.

A starting point could be a `setMediaType(string)` function defined on Interactions of `ExposedThings`. It would provide the Runtime with information what `forms` data to put into the TD. I chose a function, as `mediaType` is not a member of Interactions, but Forms in the TD; thus, it is not part of the TD model, but a function that augments the model within the Scripting API (compare for instance `.invoke()` for Actions).

zolkis:
Fixed by #113.

egekorkan:
I cannot see how this fixed in the current spec. Also a variation of this issue would be what would happen in the case where there are two forms with different media types. In this case, setMediaType() would mean one possible media type per interaction.

Issue #111, opened 2018-04-18
What we want from the Scripting API when based on Simplified TD


mkovatsc:
The Thing Description (TD) uses a formal, and hence well-defined model that describes Things. Among other details, Things can have metadata attributes such as `name` and `id`, semantic metadata attributes such as `@context` and `@type` (obviously using JSON-LD keywords), interaction definitions in the structured attributes `properties`, `actions`, and `events`, and finally links to other Things or metadata documents in the structured attribute `links`.

This formal model and instances of the descriptions are the basis to instantiate software objects in a JavaScript runtime that act as object API to interact with the corresponding Thing. Since not all Things are the same, the object API needs to provide *introspection*, to check the values of the metadata attributes and which interactions are available. It would be most intuitive, when the software object has the same structure as the TD model. Most of this can easily be defined in WebIDL, as we have a list of identifiers.

**Problem 1:** the semantic attribute identifiers contain an `@` character, and hence are no valid identifiers for JavaScript. The solution should be to access them with the bracket notation, e.g., `myThing["@type"]` to introspect the semantic attribute `@type`. To my understanding, this would be done using a [`readonly maplike<key_type, value_type>;`](https://heycam.github.io/webidl/#idl-maplike) definition for the WebIDL `interface`. Others say "named getter" would do the trick. Please clarify. What we lose is the definition of which semantic attributes are to be expected (e.g., `@context` and `@type`). However, semantic attributes were made purely optional based on community feedback, so `maplike` should be fine, no?

Furthermore, the Thing Description must be extensible to be able to evolve other time and to adapt to the many different IoT ecosystems. For this, we allow domain-specific vocabularies, which can easily be imported in the TD document through the JSON-LD `@context` keyword. This would add additional semantic attributes that usually would be of the form `prefix:term` or could be a full IRI. The colon (`:`) also makes them invalid JavaScript identifiers, and hence we also need `maplike` for them.

**Problem 2:** There is a perceived risk of a security leak for some, as the WebIDL does not restrict what attributes will be part of the object API. Yet I fail to see the actual risk here: TDs are well defined and can be validated through multiple mechanisms such as JSON Schema, Shapes/SHACL, or even custom validators of the runtime that also knows which feature exactly is supported. The Scripting API can define a MUST be validated before instantiating the API object. Furthermore, the maplike mechanism defines which types are allowed: `key_type` is always a string, `value_type` is always a JSON data type -- never a function/callback/Promise/etc.

Summary:
[...]

Issue #110, opened 2018-04-16
Specify result of writeProperty() method


danielpeintner:
The specification does not clearly define the result of a [writeProperty()](https://www.w3.org/TR/wot-scripting-api/#the-writeproperty-method) method.

Does 
> return the result

mean returning the the set value in case of success or the HTTP status code of a PUT for example?

Note: the set value might be different to the one given in `writeProperty()` due to write property handler et cetera

zolkis:
Well, none of the methods are clearly defined yet. However, the next sentence for `writeProperty()` is
`Returns a Promise that resolves on success or rejects with an Error.`
That implies the promise is either resolved (no value specified), or rejects with an error. I guess we can remove the `"return the result"` part.

danielpeintner:
@zolkis  your comment is meant to align the WebIDL definition  for `writeProperty()`

` Promise<void>    writeProperty(DOMString name, any value);`

with the actual description in prose (which is good) by removing the return value.

I think resolving this inconsistency of  `writeProperty()` should be done differnetly by defining a return value.
Let's use for example HTTP PUT which returns in case of success various response codes (200, 204 etc). We would loose this information given that the promise can just return with no content.

Do you see what I mean?

danielpeintner:
I have been thinking that many protocol bindings might not have similar HTTP/CoAP response codes (such as MQTT).

Maybe its safer to stick to no result as you proposed and align WebIDL and prose.

Sorry for going back and forth on this issue.

zolkis:
Well, I think we need to align this with Protocol Bindings and Security TF. I am not sure whether in principle should we expose underlying protocol details in this case, but technically it would be easy resolving with a value or an object that exposes details even on a successful WoT operation (if available). 

In the particular case of property update, we might not need that, but in general we should have guidelines for it.

zolkis:
This is still valid.

zolkis:
As discussed in https://www.w3.org/2018/09/17-wot-minutes.html,
currently the WebIDL definition of [`write()`](https://w3c.github.io/wot-scripting-api/#the-thingproperty-interface) specifies `void` data returned in the `Promise`, but implementations MAY choose to include error codes from underlying protocols, if the security considerations for the given application allow that. Clients can check on the existence of the provided data and act accordingly.

So the WebIDL should be changed to return `Promise<any>` (permissive version) or `Promise<long>` (hardened version) and text that explains this should return either no data, or an error code if allowed.

[...]

Issue #109, opened 2018-04-10
Refineing the WoT.discover() function


mkovatsc:
We currently have `Observable<ConsumedThing> discover(optional ThingFilter filter);` to dynamically discover Things. We should spend some time on improvement for the Korea WD release.

The optional filter looks like this:
```
dictionary ThingFilter {
    DiscoveryMethod      method = "any";
    USVString            url;
    USVString            query;
    sequence<Dictionary> constraints;
};
```

The possible methods are:
* "any" does not provide any restriction
* "local" for discovering Things defined in the same device
* "nearby" for discovering Things nearby the device, e.g. by Bluetooth or NFC
* "directory" for discovery based on a service provided by a directory or repository of Things
* "broadcast" for an open ended discovery based on sending a request to a broadcast address
* "other" for a proprietary method defined by the solution.

mkovatsc:
It might be better to return TDs by the Observable instead of instances of `ConsumedThing`: The script has more control over how many instances are created; `discover()` becomes the complement of the manual `fetch()`.

mkovatsc:
What is the difference between "nearby" and "broadcast"?

"nearby" should probably talk about proximity or distance, as Bluetooth and NFC are examples for broadcast. Even rename to "proximity"?

zolkis:
"nearby" is meant for the likes of Bluetooth and NFC. "broadcast" is for the case when a broadcast request is sent (like in OCF endpoint discovery).

Returning TDs instead of ConsumedThing instances makes a lot of sense to me.

mkovatsc:
My point is about tweaking the terms to be more precise and intuitive given other usage. Technically Bluetooth is broadcasting (radio is always a broadcast channel).

BLE and some 802.15.4 radios support distance estimations and NFC is meant for ultimate proximity, which is why I think "proximity" might be better than "nearby", as the latter is very vague. We could also talk about an optional distance threshold parameter for the filter.

"multicast" might be the better term for the current "broadcast", as multicast discovery is the term commonly used, e.g., in CoRE and OCF.

zolkis:
Using "broadcast" for Bluetooth would be too... broad. Then it would be more clear and useful to use `"nfc"` and `"ble"`. Also, an optional timeout would also make sense for discovery.
So the updated API would look like:

```javascript
partial interface WoT {
    Observable<ThingDescription> discover(optional ThingFilter filter);
};

[...]

PR #113, opened 2018-04-30
WiP: API update

zolkis:
Work in progress. 
A series of commits for changing the API as agreed in the Prague F2F, and the following issues:
#96  #108 #109 #110 #111 #112.

### DIFF
--- a/index.html
+++ b/index.html
@@ -227,22 +227,104 @@
       // [SecureContext]
       // [NamespaceObject]
       interface WoT {
-        Observable&lt;ConsumedThing&gt; discover(optional ThingFilter filter);
+        Observable&lt;ThingDescription&gt; discover(optional ThingFilter filter);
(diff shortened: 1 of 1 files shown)


mkovatsc:
Link for convenience: https://rawgit.com/zolkis/wot-scripting-api/master/index.html

zolkis:
@danielpeintner please check [this](https://github.com/w3c/wot-scripting-api/pull/113/commits/109e49f1cdd49191c1f246208e174d65f141b63b) commit...

zolkis:
Note that there are a few small differences between the TD and Scripting, especially in DataSchema vs ValueType. For instance in Scripting, object properties contain the information whether they are required or not (by default not required), whereas in the TD there is a separate property `required` as a list of strings. Traditionally each is correct in their own context (scripting dictionaries vs JSON-LD). However, this is an easy and exact mapping for the implementations.

PR #138, opened 2018-09-03
Simplifications

zolkis:
As discussed on the Scripting call of 3. Sep.,
- Remove ThingInstance, IOW merge it to ConsumedThing and ExposedThing
- Add algorithms for ExposedThing.expose() and destroy().

### DIFF
--- a/index.html
+++ b/index.html
@@ -403,7 +403,7 @@
             If the argument <var>url</var> is not a URL, reject <var>promise</var> with <code>TypeError</code> and terminate these steps.
           </li>
           <li>
-            Make a request to fetch the content of <var>url</var> as described by the <a>Protocol Bindings</a> and wait for the reply. Implementations encapsulate the fetching process and the accepted media types (such as <code>application/td+json</code>), as far as a valid <a>Thing Description</a> can be obtained as defined in [[!WOT-TD]]. Let <var>td</var> be the <a>Thing Description</a> string-serialized from the returned content, as specified in the <a href="https://w3c.github.io/wot-thing-description/#sec-td-serialization">Thing Description serialization</a>.
+            Make a request to fetch the content of <var>url</var> as described by the <a>Protocol Bindings</a> and wait for the reply. Implementations encapsulate the fetching process and the accepted media types (such as <code>application/td+json</code>), as far as a valid <a>Thing Description</a> can be obtained as defined in [[!WOT-TD]]. Let <var>td</var> be the <a>Thing Description</a> string-serialized from the returned content, as specified in the <a data-cite="!WOT-TD#sec-td-serialization">Thing Description serialization</a>.
           </li>
           <li>
             If there was an error during the request, reject <var>promise</var> with an <code>Error</code> object <var>error</var> with <code>error.message</code> set to the error code seen by the <a>Protocol Bindings</a> and terminate these steps.
@@ -503,7 +503,7 @@
         The <dfn>TD parsing algorithm</dfn> takes a string <var>td</var> as argument and runs the following steps:
         <ol>
           <li>
-            Parse <var>td</var> according to the <a href="https://w3c.github.io/wot-thing-description/#sec-td-serialization">WoT Thing Description</a> in order to produce a <a href="https://www.w3.org/TR/json-ld/#dfn-json-object">JSON object</a> <var>json</var>. Update <var>thing</var> with the properties and values defined in <var>json</var>.
+            Parse <var>td</var> according to the <a data-cite="!WOT-TD#sec-td-serialization">WoT Thing Description</a> in order to produce a <a href="https://www.w3.org/TR/json-ld/#dfn-json-object">JSON object</a> <var>json</var>. Update <var>thing</var> with the properties and values defined in <var>json</var>.
           </li>
           <li>
             If there was an error during the parsing, throw that error and terminate these steps.
@@ -1043,7 +1043,7 @@
   <section data-dfn-for="TD-structures">
     <h2>Data types and structures</h2>
     <p>
-      The [[!WOT-TD]] specification defines the <a href="https://w3c.github.io/wot-thing-description/#sec-vocabulary-definition">WoT information model</a>, i.e. the data types and data structures used in <a>WoT Interactions</a>. In this API these definitions translate to dictionary objects that are extended with methods by the interfaces defined in this specification.
+      The [[!WOT-TD]] specification defines the <a data-cite="!WOT-TD#sec-vocabulary-definition">WoT information model</a>, i.e. the data types and data structures used in <a>WoT Interactions</a>. In this API these definitions translate to dictionary objects that are extended with methods by the interfaces defined in this specification.
     </p>
     <p>
       In order to avoid duplication of definitions, references to these data types and structures is defined in this section, but for their full description please refer to the <a href="">Thing Description specification</a>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -462,13 +468,36 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
...
+            Create an <a>ExposedThing</a> object <var>thing</var> initialized from <var>model</var>.
+          </li>
+          <li>
+            For each <a>ThingFragment</a> property, initialize the property based on the provided initial or default values provided to the local <a>WoT Runtime</a> implementation, for instance initialize:
+            <ol>
+              <li>
+                 the <code>id</code> property to be the final unique identifier of the <a>Thing</a>,
+              </li>
+              <li>
+                the <code>security</code> object to represent the actual security scheme andits properties as set up by the implementation,
```
danielpeintner: Typo: change "andits" to "and its"

zolkis:
I would also like to simplify value type definitions, as indicated in #142

zolkis:
Now SecurityScheme, DataSchema (and DataType) are "outsource" to the TD spec, and the Scripting spec contains only the runtime related aspects (such as value matching algorithm, instantiation of properties of type DataSchema and SecurityScheme).

index.html
```diff
@@ -378,7 +378,7 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
         </ol>
       </p>
     </section> <!-- Discovery -->
-
+<
     <section> <h3> The <dfn>fetch()</dfn> method</h3>
       <p>
         Accepts an <code>url</code> argument of type <code>USVString</code> that represents a URL (e.g. <code>"file://..."</code> or <code>"https://..."</code>) and returns a <code><a>Promise</a></code> that resolves with a <a>ThingDescription</a> (a serialized <a>JSON-LD</a> document of type <code>USVString</code>).
```
mkovatsc: Probably should talk about the `application/td+json` media type.
Could have a note that future media types can identify other TD serializations (e.g., CBOR).

index.html
```diff
@@ -425,10 +425,16 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
...
+          <li>
+            Create a <a>ConsumedThing</a> object <var>thing</var> initialized from <var>model</var>.
           </li>
           <li>
-            Create a <a>ConsumedThing</a> object <var>thing</var> initialized from <var>instance</var>.
+            Add the <code>read()</code> and <code>write()</code> methods to the <a>ThingProperty</a> elements so that they make requests to access the remote <a>Thing</a>s and wait for the reply, as defined by the <a>Protocol Bindings</a>.
+          </li>
+          <li>
+            Add the <code>invoke()</code> methods to the <a>ThingAction</a> elements so that they make requests to the remote <a>Thing</a> to invoke its actions, as defined by the <a>Protocol Bindings</a>.
           </li>
```
mkovatsc: Need another item for ThingEvent and subscribe()

index.html
```diff
@@ -425,10 +425,16 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
             Let <var>model</var> be the result of running the <a>TD parsing algorithm</a> with <var>td</var> as argument. If that throws an error, re-throw the error and terminate these steps.
```
mkovatsc: "model" is later used for either TD string or ThingFragment object.
Thus, I would call this here "fragment", as it is an object. The "TD parsing algorithm" could already take care of the default values (see also below).

index.html
```diff
@@ -462,13 +468,36 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
             If <var>model</var> is not an object, throw <code>TypeError</code> and terminate these steps.
```
mkovatsc: line 448/454: an ExposedThing produced from TD also offers addProperty() etc., so "adding" can also apply here. Conversely, a ThingFragment can also contain Interaction definitions.

mkovatsc: line 453/459: produce() does not take a URL. Argument should be "model".

index.html
```diff
@@ -462,13 +468,36 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
...
           </li>
           <li>
-            Let <var>instance</var> be the result of running the <a>Thing instantiation algorithm</a> with <var>model passed as parameter.
+            If <var>model</var> does not have an own property that is defined in <a>ThingFragment</a> with a default value, add that property and value to <var>model</var>.
```
mkovatsc: Check and align with "TD parsing algorithm" definition (see comment above).

index.html
```diff
@@ -425,10 +425,16 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
             Let <var>model</var> be the result of running the <a>TD parsing algorithm</a> with <var>td</var> as argument. If that throws an error, re-throw the error and terminate these steps.
           </li>
           <li>
-            Let <var>instance</var> be the result of running the <a>Thing instantiation algorithm</a> with <var>model</var> as argument.
+            If <var>model</var> does not have an own property that is defined in <a>ThingFragment</a> with a default value, add that property and value to <var>model</var>.
```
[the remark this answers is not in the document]
mkovatsc: I now see that procude() can take an object that would need the same curation step.
Maybe the "TD parsing algorithm" should be split into JSON.parse() of the string and then a validation/curation algorithm that can be reused in both consume() and produce().

index.html
```diff
@@ -462,13 +468,36 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
...
+            For each <a>ThingFragment</a> property, initialize the property based on the provided initial or default values provided to the local <a>WoT Runtime</a> implementation, for instance initialize:
+            <ol>
+              <li>
+                 the <code>id</code> property to be the final unique identifier of the <a>Thing</a>,
+              </li>
+              <li>
+                the <code>security</code> object of type <a>SecurityScheme</a> to represent the actual security scheme and its properties as set up by the implementation,
+              </li>
+              <li>
+                the <code>properties</code> property to be an array of <a>ThingProperty</a> objects in which the <code>read()</code> and <code>write()</code> methods are provided to define local methods to get and set the <a>Property</a> values,
```
mkovatsc: "**array** of <a>ThingProperty</a> objects"?
In the TD model, it is a map and most implementations would probably use a map for fast lookup. Yet this is an implementation detail.
I think we can skip this part and talk about adding the read/write methods similar to ConsumendThing, only that the access is local.
Initializing the state that holds the Property values should maybe go before adding the read and write?

index.html
```diff
@@ -703,12 +712,52 @@ <h2>The <dfn>ExposedThing</dfn> interface</h2>
...
+        The <code>expose()</code> method MUST run the following steps:
+        <ol>
+          <li>
+            Return a <code>Promise</code> <var>promise</var> and execute the next steps <a>in parallel</a>.
+          </li>
+          <li>
+            If invoking <code>expose()</code> is not allowed for the current scripting context for security reasons, reject <var>promise</var> with <code>SecurityError</code> and terminate these steps.
+          </li>
+          <li>
+            Make a request to the underlying platform to create and configure serving external requests for WoT interactions, based on the <a>Protocol Bindings</a>.
```
mkovatsc: "create and configure serving external requests" -- in particular "create" sounds confusing.
maybe: """
Make a request to the underlying platform to attach protocol listeners to the WoT interactions, based on the <a>Protocol Bindings</a>.
"""

index.html
```diff
@@ -1611,175 +1649,31 @@ <h3>The <dfn data-dfn-for="">Observable</dfn> interface</h3>
...
+        <li>
+          Minimize scripts functionality and complexity.
+        </li>
+        <li>
+          Avoid heavy functional processing in scripts without authentication.
+        </li>
+      </ol>
+    </p>
+    <p>
+      Security metadata is represented as dictionary objects of type <dfn>SecurityScheme</dfn> whose properties and sub-classes are defined in the <a href="https://w3c.github.io/wot-thing-description/#sec-security-vocabulary-definition">Thing Description specification</a>.
```
mkovatsc: This is quite lost here now. I recommend to move it to "Data types used in a Thing" like the other structures.

index.html
```diff
@@ -378,7 +378,7 @@ <h4>The <dfn>ThingFilter</dfn> dictionary</h4>
         </ol>
       </p>
     </section> <!-- Discovery -->
-
+<
```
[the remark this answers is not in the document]
mkovatsc: Was actually a problem with https://zolkis.github.io/wot-scripting-api/
But still does not belong there...

zolkis:
Thanks @mkovatsc 
I will address these comments, but not before the WG call. 
Since the changes are not done, this will postpone publication of Scripting API by one week.

danielpeintner:
Besides updating some links like "[[!WOT_TD]]" which could simply point to "#bib-wot-td" at the moment it looks good to me.

One more issue that come up recently is that besides the existing types such as "BooleanSchema" or "NumberSchema" we might also have "NullSchema".
see 
https://github.com/w3c/wot-thing-description/issues/222#issuecomment-418872522

zolkis:
Thanks Daniel.

> Besides updating some links like "[[!WOT_TD]]" which could simply point to "#bib-wot-td" at the moment it looks good to me.

I thought using explicit biblio references over simple links was encouraged in specs. If we use links, it could be a direct link, IMHO it's more convenient.

> we might also have "NullSchema".

Currently it has no reference in the TD doc, so I won't include it yet. As soon as it's published, we can link it.

zolkis:
Ah, I noticed the WOT_TD (with underscore), sorry for missing that. I have fixed/amended that. Can we merge?

zolkis:
Merging for easier review, and based on the [conditional lgtm](https://github.com/w3c/wot-scripting-api/pull/138#issuecomment-419002356).

PR #145, opened 2018-09-24
Fix #144: add setActionHandler()

### LINKED ISSUE #144: Missing setActionHandler()
`setActionHandler()` was removed as the handler can now be passed in `addAction()`.

However, if an ExposedThing is produced based on a template that already contains Action descriptions, the script must be able to attach handlers afterwards.

To fix this, the spec must re-add the `setActionHandler()` definition as before:

```
ExposedThing setActionHandler(DOMString name, ActionHandler handler);
```
```
<section> <h3>The <dfn>setActionHandler()</dfn> method</h3>
      <p>
        Takes <code>name</code> as string argument and <code>handler</code> as argument of type <a>ActionHandler</a>. Sets the handler function for the specified <a>Action</a> matched by <code>name</code>. Throws on error. Returns a reference to the same object for supporting chaining.
      </p>
      <p>
        If provided, this callback function will implement invoking an <a>Action</a> and SHOULD be called by implementations when a request for invoking a <a>Action</a> is received from the underlying platform. The callback will receive a <code>parameters</code> dictionary argument.
      </p>
      <p>
        There SHOULD be exactly one handler for any given <a>Action</a>. If no handler is initialized for any given <a>Action</a>, implementations SHOULD return error if the action is invoked by any client.
      </p>
</section>
```

@zolkis I think we should fix this ASAP before publication. I only noticed this now when updating the implementation.

zolkis:

### DIFF
--- a/index.html
+++ b/index.html
@@ -684,6 +684,7 @@
         // define Actions
         ExposedThing addAction(DOMString name, ActionFragment init, ActionHandler action);
         ExposedThing removeAction(DOMString name);
+        ExposedThing setActionHandler(DOMString name, ActionHandler action);
         // define Events
         ExposedThing addEvent(DOMString name, EventFragment event);
         ExposedThing removeEvent(DOMString name);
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -880,6 +881,35 @@ <h3>The <dfn>ActionHandler</dfn> callback</h3>
...
+        Takes <code>name</code> as string argument and <code>action</code> as argument of type <a>ActionHandler</a>. Sets the handler function for the specified <a>Action</a> matched by <code>name</code>. Throws on error. Returns a reference to the same object for supporting chaining.
+      </p>
+      <p>
+         The <code>action</code> callback function will implement an <a>Action</a> and SHOULD be called by implementations when a request for invoking the <a>Action</a> is received from the underlying platform.
+      </p>
+      <p>
+        There SHOULD be at most one handler for any given <a>Action</a> and newly added handlers replace the old handlers.
+      </p>
+      <p>
+        When an external request for invoking the <a>Action</a> identified by <var>name</var> is received, the runtime SHOULD execute the following steps:
```
mkovatsc: We probably should move this somewhere else in the next iteration.
A call can also come when only `addAction()` was used and the same applies.

mkovatsc:
Thanks!

index.html
zolkis: Right, I was thinking the same. It also applies to read, write and observe requests.

PR #148, opened 2018-10-08
Updating the content of security & privacy section

ereshetova:
This is the initial text for the Security & Privacy section for your comments with normative SHOULD statements.

### DIFF
--- a/index.html
+++ b/index.html
@@ -77,6 +77,24 @@
             publisher: "W3C",
             date: "6 February 2017"
           },
+          "WOT-SECURITY-CONSIDERATIONS" : {
+            href: "https://w3c.github.io/wot-security/",
+            title: "Web of Things Security and Privacy Considerations", 
+            publisher: "W3C",
+            date: "28 August 2017"
+          }, 
+          "WOT-SECURITY-BEST-PRACTICES" : {
+            href: "https://github.com/w3c/wot-security/blob/master/wot-security-best-practices.md",
+            title: "Web of Things Security and Privacy Best Practices (WIP)", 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1453,33 +1471,143 @@ <h3>The <dfn data-dfn-for="">Observable</dfn> interface</h3>
...
+  <ul>
+        <li>A typical way to compromise any process is to send it a corrupted input via one of the exposed
+           interfaces. 
+          <span class="rfc2119-assertion" id="scripting-sec-inputs">
+           Therefore developers SHOULD perform validation on all WoT script inputs, including 
+           <a href="https://en.wikipedia.org/wiki/Fuzzing">fuzzing</a>.
+          </span>
+            There are many tool and techniques in existence to do such validation. 
+            More details can be found in [[!WOT-SECURITY-TESTING]].
+
```
zolkis: I am not sure we should put the burden solely on script developers. We should identify the inputs that are possible via scripts and insert normative and informative instructions to WoT Runtime implementors on what and how to check.

In our case I think the TD (provided to `produce()`) is one of the possible threats. Fortunately the TD document should contain normative instructions on TD parsing, and we could hopefully link those from the Scripting spec.

Other input comes from protocol replies, and that is clearly the WoT Runtimes responsibility.

ereshetova: Sure, more we can provide is better. But there is always a part of negative testing that cannot be specified, such as fuzzing. You can build a custom fuzzer to take into account the TD format, but its inputs are not predefined.

zolkis:
Thank you, looks good! Merging.

PR #150, opened 2018-10-15
Update WebIDL

zolkis:
following ReSpec update to support newest Web IDL specification


danielpeintner:
I checked your changes and I get an error free re-spec rendering again.

+1 for merging

> In the long term we need to revisit all of WebIDL, to update to newest WebIDL idioms.

In the long term i hope we get less issues with TypeScript

PR #155, opened 2018-11-13
Rewriting S&P section without normative statements

ereshetova:
Please do not merge until everyone has commented. 
If possible, we would like to discuss this PR at the next scripting call on Monday.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1480,133 +1480,175 @@
   A detailed discussion of security and privacy considerations for the Web of Things,
   including a threat model that can be adapted to various circumstances, is
   presented in the informative document [[!WOT-SECURITY-CONSIDERATIONS]].
-  This section includes only normative recommendations relevant to
-  the WoT Thing Description.
-  </p>
+  This section discusses only security and privacy risks and possible mitigations
+  directly relevant to the scripts and WoT Scripting API.
+</p>
 
-  <p><span class="rfc2119-assertion" id="td-sec-best-practices">
+  <p>
   When designing new devices and services for use with the WoT,
-  we have documented a set of best practices
-  in [[!WOT-SECURITY-BEST-PRACTICES]] that SHOULD be followed.</span>
+  we have documented a suggested set of best practices
+  in [[!WOT-SECURITY-BEST-PRACTICES]] to improve security.
   This best-practices document may be updated as security measures evolve.
   Following these practices does not guarantee security,
   but it at least will help to avoid common known vulnerabilities and pitfalls.
   </p>
 
+  <p>
+  In addition to describing security and privacy risks relevant for
+  script developers in <a href="sec-security-consideration-script">
+  Scripts Security and Privacy Risks</a>, <a href="sec-security-consideration-runtime">
+  WoT Scripting Runtime Security and Privacy Risks</a> also presets the risks relevant
+  for the WoT Scripting Runtime itself, since both are important to take into
+  consideration while building an overall WoT software stack.
+  A suggested mitigation for each risk is also proposed.
+  </p>
 
-  <p>Below are specific recommendations related to WoT runtime implementations:
-  <ul>
-      <li>In basic WoT setups, all scripts running inside the WoT runtime are considered trusted, and
-      therefore there is no strong need to perform strict isolation between each running script instance.
-      However, depending on device capabilities and deployment use case scenario risk level it might be
-      desirable to do so.
-      <ul>
-      <li>For example, if one script handles sensitive privacy-related data and well-audited,
-      it might be desirable to separate it from the rest of the script instances to minimize the risk of data
-      exposure in case some other script inside WoT gets compromised during the runtime.
-          <span class="rfc2119-assertion" id="scripting-sec-isolation-scripts">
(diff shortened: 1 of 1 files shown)


ereshetova:
@mmccool, @zolkis, @mkovatsc  for your review

zolkis:
@danielpeintner please take a look from implementation point of view

ereshetova:
@mmccool @zolkis I have updated the section splitting the WoT Runtime & WoT Scripting risks into two subsections. Let's discuss the rest of the things (i.e. runtime & provisioning description place in the main call)

ereshetova:
@zolkis btw, any idea why ipr fails? I am signing correctly now...

zolkis:
> @zolkis btw, any idea why ipr fails? I am signing correctly now...

@ereshetova 
Click on the [details](https://labs.w3.org/repo-manager/pr/id/w3c/wot-scripting-api/155) link while signed in to github and make sure your github account is connected with your W3C account. 
Of course, commits need to be signed as well and your SSH public key uploaded to github.
If all done this way, then perhaps @ashimura could check if you're added to the group.

ereshetova:
ok, I have made an explicit link now via W3C (strangely I was sure it is pesent since I have contributed to scripting api spec just a few weeks ago without problems). Hopefully this is now resolved.

danielpeintner:
I do not see any problem from implementation point of view. Having said that, I am not a security expert either.

What I noticed was the *new* term "WoT script".
Do we need to define it. Can we simply say script?
It is not used outside Section 9.

BTW, something very minor I noticed was that sometimes WoT script is lowercase and in other places capitalized (WoT Script).

ereshetova:
Thank you Daniel for the feedback! I have now fixed the terms to just use "script" to refer to a WoT script. 
So it is hopefully good now term-wise.

zolkis:
I write here a summary of (my?) arguments from this weeks [Scripting call](https://www.w3.org/2018/11/28-wot-script-minutes.html) and [WG call]( https://www.w3.org/2018/11/28-wot-minutes.html).

1. Terms
a) The WoT Runtime (a.k.a. servient) is the SW stack that implements the WoT [network] interfaces.
b) An optional part of the WoT Runtime is the WoT Scripting Runtime, for instance a JavaScript or Python or Lua runtime.
c) A part of the Scripting Runtime is the implementation of the WoT Scripting API.

2. Scope
The WoT Scripting API spec primarily targets implementors of the API (who select/integrate/use a scripting runtime and system APIs to achieve that). The Scripting spec contains sections that also target developers who write scripts, but IMHO the Security section strictly targets the API implementors, i.e. we should not count on "cooperative security", only on enforcing security.

3. Splitting specification prose
- IMHO what belongs to the optional Scripting spec (normative when opted in) is the text belonging to what is needed to implement the Scripting API (1c from above).
- IMHO the text belonging to 1b) could go either to a separate WoT Runtime Implementation Notes doc, or into the WoT Architecture doc (because it describes the servient), or into the Scripting doc. For now, we need to check which place is better, the Architecture doc or Scripting doc.
- IMHO the text belonging to 1c) should go into the WoT Architecture doc.

4. Suggestions
- Leave this PR open as it is.
- Please create a PR on the Architecture doc and include all text referring to the runtime (WoT Runtime or Scripting Runtime).
- We discuss the final split on the next security call.

ereshetova:
OK, since the talks has happened and conclusion is that architecture document does not want to host the runtime text, I would still propose that this text is merged as part of scripting. I have updated the intro text on security to outline why we include runtime risks and mitigations there.

ereshetova:
@zolkis please review

index.html
```diff
@@ -1480,133 +1480,175 @@ <h3>The <dfn data-dfn-for="">Observable</dfn> interface</h3>
...
-            More details can be found in [[!WOT-SECURITY-TESTING]].
+  <section id="sec-security-consideration-script-input">
+   <h5>Corrupted Input Security and Privacy Risk</h5>
+   <p>
+      A typical way to compromise any process is to send it a corrupted input via one of the exposed
+      interfaces. This can be done to a script instance using WoT interface it exposes.
+    </p>
+    <dl><dt>Mitigation:</dt><dd>
+        Developers should perform validation on all script inputs.
+        In addition to input validation,
```
zolkis: This is fine to mention, but we'd need to group separately which statements target 1) API implementors, 2) script developers. 

The Scripting API spec primarily targets the API implementors (using the Runtime) and at least some of the security statements should be normative text.

The recommendations for developers should be informative text and would be preferably grouped separately.

ereshetova: OK, let's discuss this at the next's week security call, if you can make it?

zolkis: Or it might be simpler to merge this PR and then I can modify it. This way all versions will be recorded in history and you lose less time shuffling things around following other people's ideas. 
Then of course we can discuss the modified version PR in the next security call.

ereshetova: That would certainly work for me! Thank you Zoltan!

zolkis:
Merging now and will make potential changes in a subsequent PR.
Thank you for the contribution!

PR #164, opened 2019-03-15
Fix #163

### LINKED ISSUE #163: typescript definitions: Items should not be required in an ArraySchema
According to the TD spec, "Items" is not a mandatory property of an ArraySchema, but in the current typescript definitions, it is defined as mandatory.

in `index.d.ts` :
```typescript
export interface ArraySchema extends BaseSchema {
    type: "array";
    items: DataSchema;
    minItems?: number;
    maxItems?: number;
}
```

should be:

```typescript
export interface ArraySchema extends BaseSchema {
    type: "array";
    items?: DataSchema;
    minItems?: number;
    maxItems?: number;
}
```

hasbel:
Change to the typescript definitions to make "items" in ArraySchema and "properties" in ObjectSchema non mandatory (as described by the TD spec).

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -383,13 +383,13 @@ export interface StringSchema extends BaseSchema {
 
 export interface ObjectSchema extends BaseSchema {
     type: "object";
-    properties: { [key: string]: DataSchema };
+    properties?: { [key: string]: DataSchema };
     required?: Array<string>;
 }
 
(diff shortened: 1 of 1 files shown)


dontcallmedom:
Marked as non substantive for IPR from ash-nazg.

zolkis:
Thanks! NB: that part is going away now.

Issue #173, opened 2019-06-10
Separate the client and server interfaces


zolkis:
As discussed on the Munich F2F, it would make sense to split the Scripting API into:
1. client API (ConsumedThing)
2. server API (ExposedThing and eventually ThingBuilder)
3. discovery API (ThingDiscovery), which is a special client API.

danielpeintner:
+1

the `WoT` object is already now technically speaking just a *entry* point for consuming, producing and discovering a thing.

PR #174, opened 2019-06-12
Separate client and server API

zolkis:
Fixing #173 
Other small fixes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -115,13 +115,13 @@
 
   <section id="abstract">
     <p>
-      The key <a>Web of Things</a> (WoT) concepts are described in the <a href="https://w3c.github.io/wot-architecture/">WoT Architecture</a> document. The Web of Things is made of entities (<a>Thing</a>s) that can describe their capabilities in a machine-interpretable format, the <a>Thing Description</a> (TD) and expose these capabilities through the <a>WoT Interface</a>, that is, network interactions modeled as <a>Properties</a> (for reading and writing values), <a>Action</a>s (to execute remote procedures with or without return values) and <a>Event</a>s (for signaling notifications).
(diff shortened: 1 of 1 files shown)


zolkis:
@danielpeintner I'd need your input on the following.

We could further simplify the API, leaving just ConsumedThing, ExposedThing and ThingDiscovery as API entry points (constructible). But then we need to add the equivalent of consume() and produce() to them.

consume() could be moved inside ConsumedThing, renamed as `instantiate(td)`. That would also solve updating the object when the TD remotely changes.

produce() could be moved inside ExposedThing as the `instantiate(td)` method. `expose()` would work the same way. This would allow updating a TD then re-instantiate in the same ExposedThing.

Looks somewhat simpler than now with the 2 levels of interfaces.

danielpeintner:
another alternative I see is to keep the WoT entry point mostly as it was (with consume, produce, and maybe even discover) and add
* a return code / error if a given function is not available/supported
* or maybe also a helper call to check what functionality is supported in a given implementation

Does this makes sense?

zolkis:
Just to record the outcome of the call:
- we do away with ThingInstance
- in this version, we maintain the split. 
Feature detection is fine to be done in exception handling in Web APIs.

zolkis:
Pretty big change in the last commit, this is the simplest API so far. 
A bit stretching the limits with the ConsumedThing constructor, but the basic validation should be really fast and if all breaks, it can be omitted from it, given that TD validation is described in / belongs to the TD spec.

zolkis:
A TODO after the basic shape of the new API is accepted is to change all exception names to be in quotes.

danielpeintner:
I support the simple API approach (having ConsumedThing, ExposedThing and ThingDiscovery only).

Moreover, the clean initialisation in the constructor makes perfect sense to me.

zolkis:
Some new changes based on discussions:
- added UA conformance classes for client, server and discovery
- added back the WoT API object
- kept constructors for ConsumedThing and ExposedThing
- some small editorial changes.

zolkis:
Amended the previous commit to add to the commit message the issue it is fixing.

zolkis:
Review period is over, consensus between editors.
Further feedback will be handled in subsequent PR's.

PR #179, opened 2019-07-22
Correct Web IDL for new errors. Add ed note for iterables.

zolkis:


zolkis:
@danielpeintner @ashimura please check if there is something to add vs as discussed on the last Scripting call.

danielpeintner:
Thanks @zolkis. Changes are as discussed during scripting call.
+1 for merging

PR #185, opened 2019-08-20
Handle implementation feedback from node-wot

### LINKED ISSUE #182: readProperty() does not allow for uriVariables or other parameters
We decided in https://github.com/w3c/wot-scripting-api/issues/149 and https://github.com/w3c/wot-scripting-api/issues/166 that we do not need a specific way to convey parameters (e.g., uriVariables) to actions and properties.

I think this is correct for actions and setting properties.
`e.g., http://example.com/counter/actions/increment?step=12`
can be transformed to 
`thing.invokeAction("increment", {step: 12})`
Similarly for setting properties: e.g.,  `thing.writeProperty("temperature", {"value": 12, unit: "celsius"})` or so.

However, how can I map a temperature service that allows for defining temperature unit 
e.g., `http://example.com/service/properties/temperature?unit=celsius`
`thing.readProperty("temperature"); // How to convey the unit?`

Hence I propose to extend 
`readProperty(propertyName: string): Promise<any>;`
to
`readProperty(propertyName: string, parameter?: any): Promise<any>;`

### LINKED ISSUE #183: handlers for unobserving property/event missing
We currently have the following calls.

```
  Promise<void> observeProperty(DOMString name, WotListener listener);
  Promise<void> unobserveProperty(DOMString name);

  Promise<void> subscribeEvent(DOMString name, WotListener listener);
  Promise<void> unsubscribeEvent(DOMString name);
```

When acting as an ExposedThing one gets provided a `WotListener` which gets called on property change or event. Fine so far.

However, I do have issues to understand how *unsubscribing* is meant to work. What if multiple listeners are subscribed? Which one to remove?
I think subscribe methods need to return a handle which than gets passed for unsubscribe...

### LINKED ISSUE #184: getTD() method instead of td attribute
We should consider changing the td from readonly attribute to a method, since it is more flexible for implementations

zolkis:
Handles a couple of issues identified by the node-wot implementation (WiP):

Fix #183: add id to observe and subscribe
Fix #184: change .td to function .getThingDescription() and add the ||td|| internal slot
Fix #182: re-add InteractionOptions that contain uriVariables, define PropertyValueMap


zolkis:
@danielpeintner PTAL

index.html
```diff
@@ -503,11 +503,11 @@ <h2>The <dfn>ConsumedThing</dfn> interface</h2>
...
-        Promise&lt;void&gt; observeProperty(DOMString name, WotListener listener);
-        Promise&lt;void&gt; unobserveProperty(DOMString name);
-        Promise&lt;void&gt; subscribeEvent(DOMString name, WotListener listener);
-        Promise&lt;void&gt; unsubscribeEvent(DOMString name);
-        readonly attribute ThingDescription td;
+        Promise&lt;DOMString&gt; observeProperty(DOMString name, WotListener listener);
+        Promise&lt;void&gt; unobserveProperty(DOMString subscriptionId);
+        Promise&lt;DOMString&gt; subscribeEvent(DOMString name, WotListener listener);
+        Promise&lt;void&gt; unsubscribeEvent(DOMString subscriptionId);
+        ThingDescription td();
```
danielpeintner: tend to say `getTD()`

zolkis: Commit amended to use getThingDescription() instead of td()

danielpeintner:
looks good to me

PR #186, opened 2019-08-21
Fix #180: Clarify ExposedThing Property initialization

### LINKED ISSUE #180: WIP: Implementation experience while realizing new API in node-wot
Note: This is a evolving issue list that gets updated/modified/edited while realizing the new Scripting API in node-wot:

Issues:
* [x] A *light* ThingBuilder functionality like we used to have in `ExposedThing` is desirable
  * e.g., `addProperty("propName", {...}, "initValue")` allowed for initializing a property with an initial value before starting it. With the new API there is no such possibility. I think some basic `addProperty()`, `addAction()`, `addEvent()` methods on ExposedThing would be of of great help in this regard.
    * can be done in helper class
    * initializing of properties is a separate step though 
  * make *clearer* in spec/note that initializing values can be done before exposing a thing by adding `writeProperty()` in ExposedThing example right after `WoT.produce(...)`
* [x] readProperty() parameters/options, see https://github.com/w3c/wot-scripting-api/issues/182
* [x] use of getTD()/getThingDescription() method instead of td attribute (allows overloading, keep internal structure as TD, logging, ... more flexibility), see https://github.com/w3c/wot-scripting-api/issues/184
* [x] handler IDs for unobserving property/event, see https://github.com/w3c/wot-scripting-api/issues/183

### LINKED ISSUE #183: handlers for unobserving property/event missing

zolkis:
Handles a couple of issues identified by the node-wot implementation (WiP):

Fix #183: add id to observe and subscribe

### DIFF
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,0 +1,4 @@
+{
+    "src_file": "index.html",
+    "type": "respec"
+    }
--- a/index.html
+++ b/index.html
@@ -980,9 +980,10 @@
       <p class="note">
          Note that an existing <a>ThingDescription</a> object can be optionally modified (for instance by adding or removing elements on its <var>properties</var>, <var>actions</var> and <var>events</var> internal properties) and the resulting object can used for constructing an
          <a>ExposedThing</a> object. This is the current way of adding and
-         removing <a>Property</a>, <a>Action</a> and <a>Event</a> definitions,
-         and later adding the service handler functions.
-         This is illustrated in the <a href="#exposedthing-examples">examples</a>.
+         removing <a>Property</a>, <a>Action</a> and <a>Event</a> definitions, as illustrated in the <a href="#exposedthing-examples">examples</a>.
+      </p>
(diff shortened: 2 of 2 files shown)


danielpeintner:
+1

zolkis:
The amends have been:
- remove the old "mock" part of the ExposedThing example resulting in cleaner example
- make the clarification of initialization vs exposing as a Note.

PR #205, opened 2020-02-24
Returning values from writes

zolkis:
Captures the decision made in the Scripting call on 24.2.2020: 
- include an ed's note about write and return values
- update the design rationale file.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-scripting-api/pull/205.html" title="Last updated on Feb 24, 2020, 2:36 PM UTC (ed634fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/205/cd2f7ee...zolkis:ed634fc.html" title="Last updated on Feb 24, 2020, 2:36 PM UTC (ed634fc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -716,6 +716,15 @@
           </li>
         </ol>
       </div>
+      <p class="ednote">
+        As discussed in <a href="https://github.com/w3c/wot-scripting-api/issues/193">
+        Issue #193</a>, the design decision is that write interactions only
+        return success or error, not the written value (optionally).
+        <a>TD</a>s should capture the schema of the <a>Property</a>
+        values, including precision and alternative formats. When a return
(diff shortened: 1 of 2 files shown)


zolkis:
IIRC we created the rationale.md to capture the most important design decisions from issue discussions and calls. 

In some specs there is also an appendix, but usually (more recently) this is done in a separate document.

However, in this specific case I agree with @mmccool that a note in the spec is warranted. This note is merely informative and could go away, but the rationale.md should contain it anyway (and likely needs to be extended with other design decisions we made, including how to use write handlers, but that in a separate PR, later).

zolkis:
Can we merge this, or wait, or is something to be changed/fixed/improved?

danielpeintner:
OK with merging

Issue #208, opened 2020-03-22
Event Affordance subscription and cancellation payloads are not usable by scripting 


egekorkan:
In the TD spec, the [Event affordance](https://www.w3.org/TR/wot-thing-description/#eventaffordance) has the `subscription` and `cancellation` keys. With these keys, one can describe a payload that needs to be passed along for subscribing or unsubscribing from an event. 

In the Scripting API, this behavior is not specified explicitly for the [subscribeevent method](https://www.w3.org/TR/wot-scripting-api/#the-subscribeevent-method) and is not specified at all for the [unsubscribeevent method](https://www.w3.org/TR/wot-scripting-api/#the-unsubscribeevent-method). For `subscribeevent`, it is said that one can use `InteractionOptions` vocabulary which however says that currently they are used only for URI variables. For `unsubscribeevent`, `InteractionOptions` do not seem to be even allowed.

A further discussion would be whether these are actually interaction options or `params` like the `invokeAction()` method.

zolkis:
The question is, from where the subscription and cancellation options (data) come from: application or implementation/bindings?

If they come from the app, then:
- in subscribeEvent, the InteractionOptions should carry an optional `subscribeOptions` property (`any`), which is opaque to the implementation and used by the protocol bindings on the other end;
- in unsubscribeEvent, there should be InteractionOptions, that can carry an optional `unsubscribeOptions` (`any`) which is again opaque to the implementation and used by the protocol bindings on the other end.

If these come from the protocol bindings, then they need not be exposed to the app. What should be the behaviour if the app defines these, should the implementation try to use them (how) and signal error or ignore them?

egekorkan:
That's a very good question. Here are some TDs that define `subscription` and `cancellation` payloads:
- [From Oracle and @mlagally ](https://github.com/w3c/wot-testing/blob/master/data/testing/tests/2019-05/inputs/Oracle/oracle-simulators/Blue_Pump_Events.jsonld#L343):
This TD requires a very complex payload for subscription. I would doubt that an implementation can figure this out on its own or that such information could be specified statically in a configuration file.
- [From Siemens and @sebastiankb ](https://github.com/w3c/wot-testing/blob/master/data/testing/tests/2019-05/inputs/Siemens/siemens-customEvent/eventTest-siemens.json#L44):
[...]

Issue #201, opened 2020-01-13
Fix property interaction algorithms to correctly handle Forms, contentType and DataSchema


zolkis:
As raised in #198 and then discussed in many calls, the current interaction algorithms don't handle correctly the WoT interactions.

A WoT interaction might contain multiple [Forms](https://w3c.github.io/wot-thing-description/#form), each of which define a `contentType` and for some content types, may define `DataSchema` in the Property definition itself. 

In most cases, DataSchema will be defined, but e.g. for media types the value will be binary and the media type needs to be passed to the application (in the Form) in order to be able to parse it. But currently there is no mechanism for that.

The `readProperty()` algorithm needs to choose one of the Forms, first handle `contentType`, then if the case, handle `DataSchema` and then validate/handle the binary `value` according to these.

This issue replaces #198 as it is wider.

zolkis:
Implementations could map WoT property read interactions to a JS interface of obtaining the data (based on [Web Platform](https://en.wikipedia.org/wiki/Web_platform) idioms). An attempt has been described [here](https://github.com/w3c/wot-scripting-api/issues/198#issuecomment-547589233):

> The proper way would be a Scripting-defined Value dictionary like the one above (and below): currently we need those 4 properties in the Web Platform to represent binary or textual data with encoding and language.
So this is my proposal to handle values in WoT Scripting:

```javascript
interface Value {
  readonly attribute USVString? mediaType;
  readonly attribute USVString? encoding;
  readonly attribute USVString? lang;
  readonly attribute DataView? data;
}
```

> Then we can create/read text, JSON or binary by choosing the encoder/decoder.

egekorkan:
We can also look at the [fetch API for the browsers.](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch)  There, one can do `response.json()` in order to obtain the JSON value of the response. Maybe we can do something similar where `readproperty` returns a value and the script programmer has to parse the data out of it like implied in the previous comment(s).

Another way I can think of this would be to give `readProperty` method an optional argument, where the programmer can specify the desired contentType. The Scripting runtime would then try to convert the payload if it can, otherwise return the binary. If this argument is not given, the runtime tries to convert to JSON by default.

I think it is also worth thinking the similar case for other Interaction Affordances, i.e. choosing an Action Form based on the content type or the Event form to subscribe to.

zolkis:
[...]

PR #209, opened 2020-04-04
Add InteractionData

zolkis:
Add InteractionOptions to `unobserveProperty()` and `unsubscribeEvent()`. Addresses #201 

Add `InteractionData` definition, usage and examples. Addresses #208

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-scripting-api/pull/209.html" title="Last updated on Jun 1, 2020, 5:36 PM UTC (45b888a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/209/8037a79...zolkis:45b888a.html" title="Last updated on Jun 1, 2020, 5:36 PM UTC (45b888a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -589,13 +589,19 @@
     <section>
       <h3>The <dfn>readProperty()</dfn> method</h3>
       <div>
-        Reads a <a>Property</a> value. Takes a string argument |propertyName| and and an optional {{InteractionOptions}} |options:InteractionOptions| argument. It returns a <a>Property</a> value represented as `any` type. The method MUST run the following steps:
+        Reads a <a>Property</a> value. Takes a string argument |propertyName|
+        and an optional {{InteractionOptions}} |options:InteractionOptions|
+        argument. It returns a <a>Property</a> value represented as `any` type.
+        The method MUST run the following steps:
         <ol>
           <li>
-            Return a {{Promise}} |promise:Promise| and execute the next steps <a>in parallel</a>.
+            Return a {{Promise}} |promise:Promise| and execute the next steps
+            <a>in parallel</a>.
           </li>
           <li>
-            If invoking this method is not allowed for the current scripting context for security reasons, reject |promise| with a {{SecurityError}} and abort these steps.
+            If invoking this method is not allowed for the current scripting
+            context for security reasons, reject |promise| with a
+            {{SecurityError}} and abort these steps.
           </li>
           <li>
             Let |interaction| be the value of ||td||'s |properties|'s
@@ -608,10 +614,14 @@
             |interaction|'s |forms| whose |op| is `readproperty`.
           </li>
           <li>
-            If |form| is failure, reject |promise| with a {{SyntaxError}} and abort these steps.
+            If |form| is failure, reject |promise| with a {{SyntaxError}} and
+            abort these steps.
           </li>
           <li>
-            Make a request to the underlying platform (via the <a>Protocol Bindings</a>) to retrieve the value of the <a>Property</a>given by |propertyName| using |form| and the optional URI templates given in |options|' {{uriVariables}}.
+            Make a request to the underlying platform (via the
+            <a>Protocol Bindings</a>) to retrieve the value of the <a>Property</a>
+            given by |propertyName| using |form| and the optional URI templates
+            given in |options|' {{uriVariables}}.
           </li>
           <li>
             If the request fails, reject |promise| with the error received from the <a>Protocol Bindings</a> and abort these steps.
@@ -666,12 +676,12 @@
                 Let |value| be the value of |result|'s |key|.
               </li>
               <li>
-                Let |dataschema| be the value of ||td||'s |properties|'s |key|.
+                Let |schema| be the value of ||td||'s |properties|'s |key|.
               </li>
               <li>
                 Let |property| be the result of running
                 <a>parse interaction response</a> on |value|, |form| and
-                |dataschema|.
+                |schema|.
               </li>
             </ol>
           </li>
@@ -723,12 +733,12 @@
                 Let |value| be the value of |result|'s |key|.
               </li>
               <li>
-                Let |dataschema| be the value of ||td||'s |properties|'s |key|.
+                Let |schema| be the value of ||td||'s |properties|'s |key|.
               </li>
               <li>
                 Let |property| be the result of running
                 <a>parse interaction response</a> on |value|, |form| and
-                |dataschema|.
+                |schema|.
               </li>
             </ol>
           </li>
@@ -820,14 +830,14 @@
             `null`.
           </li>
           <li>
-            Let |dataschemas:object| be an object and for each string |name:string|
+            Let |schemas:object| be an object and for each string |name:string|
             in |propertyNames| add a property with key |name| and let its value
             be the value of ||td||'s |properties|'s |name|.
           </li>
           <li>
             For each key |key:string| in |properties|, take its value as |value|
             and run the <a>create interaction data</a> steps on |value|, |form|
-            and the value for |dataschema|'s |key|.
+            and the value for |schema|'s |key|.
             If that fails for any |name|, reject <a>promise</a> with a
             {{SyntaxError}} and abort these steps.
           </li>
@@ -942,8 +952,7 @@
             If |form| is failure, reject |promise| with a {{SyntaxError}} and abort these steps.
           </li>
           <li>
-            Make a request to the underlying platform (via the <a>Protocol Bindings</a>) to stop observing the <a>Property</a> identified by |propertyName|, with |form| and optional URI templates given in |options|'
-            {{uriVariables}}.
+            Make a request to the underlying platform (via the <a>Protocol Bindings</a>) to stop observing the <a>Property</a> identified by |propertyName|, with |form| and optional URI templates given in |options|' {{uriVariables}}.
           </li>
           <li>
             If the request fails, reject |promise| with the error received from the <a>Protocol Bindings</a> and abort these steps.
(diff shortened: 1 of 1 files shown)


danielpeintner:
I wonder whether the input to `writeProperty()` and `invokeAction()` should be of `InteractionData` also (instead of `any`).

zolkis:
> I wonder whether the input to writeProperty() and invokeAction() should be of InteractionData also (instead of any).

Yes, this is an important topic and it is not specified.

Typically a script would like to use JS structures as data source: simple types, arrays and objects.
What we need is an algorithm on how to serialize these data sources using the protocol bindings to make the interactions happen.

IMHO it makes little sense to convert JS types to binary data before feeding to the API calls. Implementations should be able to do that. The question is what and to what extent should we specify in the Scripting API.

danielpeintner:
> I wonder whether the input to `writeProperty()` and `invokeAction()` should be of `InteractionData` also (instead of `any`).

I had an offline discussion with @sebastiankb @egekorkan and @wiresio  
The general consensus was that the API should be balanced meaning that IF `readProperty()` returns `InteractionData` `writeProperty(`) should also expect `InteractionData`.

I think there is also a good use-case. Let's assume getting a property "foo" is possible via HTTP (returning JSON) and CoAP (returning CBOR). When writing the value again it is possible to pass again in JSON or CBOR. With `InteractionData` the runtime knows whether it should use HTTP/CoAP depending on the mediaType.

I will try to look at the overhead of "encoder" and "decoder".

zolkis:
> The general consensus was that the API should be balanced  

Apart from esthetic reasons, why the API should be balanced (or symmetric)?
It would be too much unnecessary burden requiring from applications to do a serialization that implementations could do, by looking at the JS type and the bindings (the default or the specified one). Every time the API is called.

> Let's assume getting a property "foo" is possible via HTTP (returning JSON) and CoAP (returning CBOR). When writing the value again it is possible to pass again in JSON or CBOR. With InteractionData the runtime knows whether it should use HTTP/CoAP depending on the mediaType.

For selecting the bindings, we already have a mechanism.
If implementations have to deduce which bindings to use based on media type, that would be similar to taking JS type and deducing binding serialization.
So we would be doing exactly the opposite that would make sense for me:
- when writing, deduce serialization from JS type and select the bindings with an index
- when reading, produce InteractionData.

danielpeintner:
> For selecting the bindings, we already have a mechanism.

This is correct, but we are shifting the complexity to the client while in this case the runtime could figure it out itself. 

What you are saying is that everything "could" be handled by the client. If we want to go along this path we should not even introduce the mechanism of `InteractionData` in the first place.

zolkis:
The idea was not to push complexity to apps, but
- whenever possible, implementations should do the hard work
- but when they cannot handle the data, they expose it to the app, together with the context.

That should be both for reading and writing.

But I agree that if we have a "content exception" mechanism when reading, we should also have a manual way to specify content when writing. In that sense I agree.

However, requiring the "exception" mechanism for every case, for the sake of uniformity is IMHO nasty with developers :).

I guess we need to experiment with this. 

We also need some statistics on what data types are used most frequently. If there is any input on that, please share.

zolkis:
One way to deal with this would to continue using `any` in the API, but the applications could check it as follows.

For Property reads, the value could be one of the following:
- `null` maps from `NullSchema`
- a _number_ maps from `NumberSchema` or `IntegerSchema`
- `false` and `true` map from `BooleanSchema`
- a _string_ maps from `StringSchema` (if no language and encoding need to be specified)
- an _array_ maps from `ArraySchema`
- if an _object_ is returned, apps SHOULD check its type; if the type is `InteractionData`, then it should be handled in the binary way using its content type, lang and encoding; otherwise, handle it as mapped from `ObjectSchema`, meaning a property bag with values as defined here (recursive definition).

For Property writes, it would be similar. If an app provides an `InteractionData` object to the methods, it SHOULD be handled as such by implementations. Otherwise an object passed to writes should map to an `ObjectSchema`. The other types would be inferred from the JS values given to the method vs the DataSchema of the Property, or if not available then the Form (selected by app or default).

Would that work for you?
@danielpeintner @egekorkan

zolkis:
I added detailed algorithms on how to handle DataSchema and contentType on reads and writes.

Also defined how to handle forms in all interactions.

danielpeintner:
I suggest merging the changes w.r.t. `InteractionOptions` and `formIndex`, errors etc and create a new issue with how to deal with `InteractionData` or checking the return value.

I am a bit unsure about all these type checks et cetera...

zolkis:
As explained [here](https://github.com/w3c/wot-scripting-api/issues/201#issuecomment-623550139), maybe the best option is to merge this PR in this form, and then discuss further or alternative options on exposing interaction data.

danielpeintner:
I agree with merging and gathering experience.

danielpeintner:
I *started* to apply the proposed updates in a WIP PR (see https://github.com/eclipse/thingweb.node-wot/pull/228).

I think there is one issue with `InteractionOptions` that should mark all entries optional with `?`.

```
dictionary InteractionOptions {
  unsigned long? formIndex;
  object? uriVariables;
};
```

Am I wrong?

zolkis:
The last commit is a big change and it reflects the understanding from today's Scripting call:
- use `ReadableStream` in `InteractionData`, both for reading and writing (one can create a `Body` for `Response` from a `ReadableStream` as well),
- included the selected `Form` object in `InteractionData`,
- From `InteractionData` removed `mediaType` and `encoding` and `lang` (since it's available from the form or from the TD),
- return `InteractionData` from reads, observe, actions and events,
- rewrite the "parse interaction data" algorithm and update its invocations,
- rewrite the "create interaction data" algorithm and update its invocations,
- update example with a reading function that uses either `value` or the stream.

index.html
```diff
@@ -1085,231 +1147,219 @@ <h2>ConsumedThing Examples</h2>
...
       Previously `any` has been used for representing data that
-      could be returned by <a>WoT Interactions</a>. Since data type can be
-      specified in the <a>TD</a> by a <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#form">Form</a>'s `contentType`
-      and optionally a <a>DataSchema</a>,
-      this interface enables handling returned data by applications.
+      could be returned by <a>WoT Interactions</a>. Data type is specified in
+      the <a>TD</a> by a <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#form">Form</a>'s `contentType`
+      and optionally a <a>DataSchema</a>.
+      This interface enables handling returned data by applications as a stream
+      or
```
danielpeintner, line 1158: There seems to be something missing

zolkis: Yes, thanks, I fixed that (reformulated the whole part, please check if it's accurate).

danielpeintner: The description looks fine now. I still need to make up my mind and explore `InteractionData` with code.

zolkis:
@relu91 PTAL :)

relu91:
See my comment in #201. Except that, it seems really good. however, I agree with @danielpeintner, I would like to have some time to experiment a little bit with it. Especially, with the algorithm defined.

zolkis:
Yes, let's see how it works out when used in code. I chose this version since it's simpler to use, however there are arguments for the async function.

zolkis:
I made the updates summarized [here](https://github.com/w3c/wot-scripting-api/issues/201#issuecomment-633674708).

Note that I had to add an internal slot for [[value]] in `InteractionData`, since consecutive invocations of `value()` should not re-parse the stream. It's also useful when writing.

zolkis:
Amended last commit with adding the forgotten `dataUsed` definition.

index.html
```diff
@@ -929,31 +1106,423 @@ <h2>ConsumedThing Examples</h2>
...
+        </li>
+        <li>
+          If the value of the <a>[[\value]]</a> <a>internal slot</a> is not
+          `undefined`, resolve |promise| with that value and abort these steps.
+        </li>
+        <li>
+          If the value of the |data| property is not a {{ReadableStream}} or if
+          |dataUsed| is `true`, or if |form| is `null` or if |schema| or its
+          |type| are `null` or `undefined`,
+          reject |promise| with {{SyntaxError}} and abort these steps.
```
relu91, line 1262: Don't we have a better error to describe this situation? like [NotReadable](https://heycam.github.io/webidl/#notreadableerror)

index.html
```diff
@@ -929,31 +1106,423 @@ <h2>ConsumedThing Examples</h2>
...
+          "https://fetch.spec.whatwg.org/#concept-read-all-bytes-from-readablestream">
+          reading all bytes</a> from |data| with |reader|.
+        </li>
+        <li>
+          Set |dataUsed| to `true`.
+        </li>
+        <li>
+          If |form|'s |contentType| is not `application/json` and if a mapping is
+          available in the <a>Protocol Bindings</a> from |form|'s |contentType|
+          to [JSON-SCHEMA], transform |bytes| with that mapping.
```
relu91, line 1281: I'd check the content type before reading the response. So that the user has still the opportunity to parse the stream if the content-type has no mapping to JSON-Schema

index.html
```diff
@@ -929,31 +1106,423 @@ <h2>ConsumedThing Examples</h2>
...
+          [[\value]] <a>internal slot</a> as its
+          <a href="https://streams.spec.whatwg.org/#underlying-source">
+          underlying source</a>.
+        </li>
+        <li>
+          Return |idata|.
+        </li>
+      </ol>
+    </section>
+
```
relu91, line 1494: Is this algorithm updated to the latest changes? I miss understanding why we parse source at the creation of `InteractionData`. Shouldn't it be done inside the `value` function?

index.html
```diff
@@ -929,31 +1106,423 @@ <h2>ConsumedThing Examples</h2>
...
+          reader.read().then(function process({ done, chunk }) {
+              if (done) {
+                value += chunk;
+                return value
+              }
+              value += chunk;
+              return reader.read().then(process);
+            });
+        };
```
relu91, line 1150: As in my last comment, is this updated with the latest changes?

index.html
zolkis: Makes sense.

index.html
zolkis: Developers can read the `form`'s `contentType` before deciding to call `value()` or use the stream.

BTW I am thinking to add a note that implementations MAY decide to duplicate the stream, i.e. `value()` would consume one, but then `data` would be still available for read. The only thing devs should do is check the `dataUsed` attribute beforehands in order to avoid errors - or then handle the errors.

index.html
zolkis: Yes. We have not really discussed writing, but it's a bit different since it takes `any` as value (as aligned with other APIs as well) and one of the values may be a prepared `InteractionData`. But it may be anything else as well, as the algorithm explores. 

If an `InteractionData` is provided, the writing algorithm doesn't use `value()`, it uses its `data`.

index.html
zolkis: No, I forgot to update the example. Fixing...

index.html
relu91: Now I got it, sorry I misinterpreted the paragraph I thought the algorithm was for creating an InteractionData from the runtime perspective! 

it makes perfect sense now!

index.html
relu91: > Developers can read the form's contentType before deciding to call value() or use the stream.

yeah, right. However, I think it does not change the algorithm so much and it gives an opportunity to handle this case only if the method fails. what do you think?

> BTW I am thinking to add a note that implementations MAY decide to duplicate the stream, i.e. value() would consume one, but then data would be still available for read. The only thing devs should do is check the dataUsed attribute beforehands in order to avoid errors - or then handle the errors.

Can you please provide a use case when you want to read the stream anyway even if you **have** already a parsed value? Aside from when the value function fails I cannot find any. Moreover, even in such a scenario, I am not sure if another stream would be valuable. Here it is a list of possible `value` errors (please correct me if I miss any) :
- malformed JSON: any other parser would fail anyway --> no need to read from a stream
- no mapping from the content type to JSONSchema: As I suggested the algorithm of the value function can be engineered to not consume the stream --> no need to have another stream
- invalid data schema: I think this is the only use case to try consuming the stream manually after value failure. However, should we encourage this practice? Invalid data schema means that something is really off...

zolkis: I just want to avoid duplicate network requests, since that is far more expensive than duplicating a stream (which boils down to pointer ownership and no data duplication necessarily).

The normal flow would be to try to use `value()`, which may fail, in which case  using `data` may be attempted.

Now I think I misunderstood your first comment - what you seem to wish is that `value()` fails as early as possible before disturbing the stream. Yes, let's try to do that first and see how it works.

index.html
zolkis: Fixed

index.html
zolkis: Also updated other errors and made algorithms re-throw where it made sense.

zolkis:
@relu91 any better now?

PR #220, opened 2020-06-18
Fix #215: make the API entry point a namespace. 

### LINKED ISSUE #215: [typescript] Use namespace instead of interface for WoT functions
Currently, the `wot-typescript-definition` file export WoT functions as part of the  WoT interface. This interface is under the namespace WoT, therefore accessible with `WoT.WoT` path. This works well for node.js environments, while it may not work as intended inside browsers or servient runtimes. 

Following the [IDL](https://w3c.github.io/wot-scripting-api/#the-wot-interface) for the `WoT` interface, those API should be accessed from the global scope as, for example, `WoT.consume(..)`. I assume that also runtime scripts (i.e. scripts run by servients) MAY use the same mechanism (instead of `import` or `require`). However, the typescript definition does not define a globally scoped variable. 

[Typescript documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html) suggests using a namespace to specify a global variable. I made a refactored version of the type-definition [here](https://github.com/UniBO-PRISMLab/wam/blob/master/templates/simplets/types/wot.d.ts). With that d.ts, vscode (and tsc) correctly identify `WoT` global variable inside node-wot scripts as shown in [this tweet](https://twitter.com/i/status/1238072300806778880)

I am not sure that this is the intended behavior, therefore I am not creating a PR. However, I'd like to understand whether my suggestion could be an improvement to the current `wot-typescript-definitions` or not.

zolkis:
Also, apply Web IDL namespace rules.

### DIFF
--- a/index.html
+++ b/index.html
@@ -369,21 +369,21 @@
     </section>
   </section>
 
-  <section data-dfn-for="WebOfThings">
-  <h2>The <dfn>WebOfThings</dfn> namespace</dfn></h2>
+  <section data-dfn-for="WOT">
+  <h2>The <dfn>WOT</dfn> namespace</dfn></h2>
   <p>
     Defines the <dfn>WoT API object</dfn> as a singleton and contains the API
     methods, grouped by conformance classes.
   </p>
   <pre class="idl">
     [SecureContext, Exposed=(Window,Worker)]
-    namespace WebOfThings {
+    namespace WOT {
       // methods defined in UA conformance classes
     };
   </pre>
   <section> <h3>The <dfn>consume()</dfn> method</h3>
     <pre class="idl">
-      partial namespace WebOfThings {
+      partial namespace WOT {
         Promise&lt;ConsumedThing&gt; consume(ThingDescription td);
       };
     </pre>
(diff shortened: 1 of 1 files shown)


zolkis:
I thought this issue was to draw attention to namespaces because of the node-wot issue.
From the other hand, IIRC we have got that feedback to use a namespace also from elsewhere.

I am fine with the WOT name, this PR was a bit provocative with this  (better readable) name.
I thought spelling it out plays better with Web IDL guidance on namespace naming, and be more clear in the same time. IMHO it makes things more clear/explicit when you import/use WebOfThings vs WOT. But no hard take on that.

zolkis:
https://heycam.github.io/webidl/#idl-namespaces

danielpeintner:
> It seems right, even if I liked more `WOT` as namespace than `WebOfThings`. However, I understand that `WebOfThings` is more evocative and less obscure. What about simply `Things`?

I agree with @relu91 
I don't think changing "WOT" to "WebOfThings" is a good idea. It conflicts with a product too.

And in general I think this is a TypeScript issue.. but it may also relate to IDL namespaces...

zolkis:
Changed the name back to WOT.

Issue #230, opened 2020-08-03
Some intial read-through comments


danielpeintner:
Find below some aspects we should discuss...

# Abstract
* link to WoT Architecture points to CR, should be TR
  (Q: Should we point to the latest document? https://www.w3.org/TR/wot-architecture/ OR to dated URI ?)
* link to TD  points to CR, should be TR
  (Q: Same as for Architecture)
* Editor's note: "plug-fests" .. use "Plugfest" everywhere else... 

# F. References

* Updates for 
  * [WOT-ARCHITECTURE] --> Link+Date
  * [WOT-PROTOCOL-BINDINGS] --> Date
  * [WOT-SECURITY-BEST-PRACTICES] Broken link
  * [WOT-SECURITY-TESTING] Broken link
  * [WOT-TD] --> Link+Date

## Minor updates/issues provided by PR
https://github.com/w3c/wot-scripting-api/pull/229

egekorkan:
Please find my review in the attached pdf. I have read until the chapter 7 and will read the rest later but I wanted to send it as soon as possible anyways.

I have put mostly notes, replace text, insert text, strikethrough and I have underlined some text that I see "suspicious", like ref errors
[ScriptingAPI-review-egekorkan.pdf](https://github.com/w3c/wot-scripting-api/files/5017322/ScriptingAPI-review-egekorkan.pdf)

danielpeintner:
# checking Ege's PDF chapters
@zolkis Chapter 1-2
@relu91  Chapter 3-5
@danielpeintner Chapter 6

danielpeintner:
@egekorkan others please provide comments in Github index.html (line by line commenting possible?)
Thanks!

egekorkan:
I was thinking of the best way to do it as well. Should I create a PR with my comments that will be never merged?

zolkis:
We can handle the PDF for now.
But in general, one can view the index.html and comment on each line.
Also, one can choose any commit and comment on that, creating new issues that refer to those comments. 
Also, a WoT WG member can create a PR with trivial or non-trivial fixes.

egekorkan:
So the problem is that I cannot comment on any file on the repository. The comment features become available when there is a PR or if I choose a certain commit and comment on that. Since it was a general review of the spec and not a single commit, I had find some other solution. Choosing a certain commit is also not good for reviewing the whole document since I cannot see the whole document on a commit that addresses a single section (most commits are like that).
[...]

PR #231, opened 2020-08-10
Minor changes on chapters 4 and 5

relu91:
This PR reports @egekorkan review changes in chapters 4 and 5 (no comments were left on chapter 3). See the related https://github.com/w3c/wot-scripting-api/issues/230#issuecomment-668142199 and the [pdf document](https://github.com/w3c/wot-scripting-api/files/5017322/ScriptingAPI-review-egekorkan.pdf)

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/231.html" title="Last updated on Aug 14, 2020, 9:16 AM UTC (8c948cf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/231/f7eef4c...relu91:8c948cf.html" title="Last updated on Aug 14, 2020, 9:16 AM UTC (8c948cf)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -354,6 +354,9 @@
             validation</a> fails on |td|, [= exception/throw =] a {{"TypeError"}} and abort these steps.
           </li>
         </ol>
+        <p class="ednote" title="Handling default values">
+          Additional steps may be added to fill the default values of mandatory fields.
+        </p>
       </div>
     </section>
   </section>
--- a/index.html
+++ b/index.html
@@ -332,7 +332,7 @@
         <a>TD</a>.
       </p>
       <div>
-        run the following steps:
+        To <dfn>expand a TD</dfn> given |td:ThingDescription|, run the following steps:
         <ol>
           <li>
             For each item in the <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#sec-default-values">TD default values</a> table from [[!WOT-TD]], if the term is not defined in |td|, add the term definition with the default value specified in [[!WOT-TD]].
@@ -347,15 +347,10 @@
         Therefore, this API expects the {{ThingDescription}} objects be validated before used as parameters. This specification defines a basic <a>TD</a> validation as follows.
       </p>
       <div>
-        run the following steps:
+        To <dfn>validate a TD</dfn> given |td:ThingDescription|, run the following steps:
         <ol>
           <li>
             If |td| is not an object, [= exception/throw =] a
-            {{"TypeError"}} and abort these steps.
-          </li>
-          <li>
-            If <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#json-schema-for-validation">JSON schema
-              validation</a> fails on |td|, [= exception/throw =] a
             {{"TypeError"}} and abort these steps.
           </li>
           <li>
@@ -363,6 +358,10 @@
             <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#thing">Thing</a>
             that don't have <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#sec-default-values">default definitions</a>
             are missing from |td|, [= exception/throw =] a
+            {{"TypeError"}} and abort these steps.
+          </li>
+          <li>
+            If <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#json-schema-for-validation">JSON schema validation</a> fails on |td|, [= exception/throw =] a
             {{"TypeError"}} and abort these steps.
           </li>
         </ol>
--- a/index.html
+++ b/index.html
@@ -350,23 +350,8 @@
         To <dfn>validate a TD</dfn> given |td:ThingDescription|, run the following steps:
         <ol>
           <li>
-            If |td| is not an object, [= exception/throw =] a
-            {{"TypeError"}} and abort these steps.
-          </li>
-          <li>
-            If any of the mandatory properties defined in [[!WOT-TD]] for
-            <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#thing">Thing</a>
-            that don't have <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#sec-default-values">default definitions</a>
-            are missing from |td|, [= exception/throw =] a
-            {{"TypeError"}} and abort these steps.
-          </li>
-          <li>
-            Fill any missing mandatory property defined in [[!WOT-TD]] that have a 
-            <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#sec-default-values">default definition</a> with its default.
-          </li>
-          <li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -332,7 +332,7 @@ <h2>The <dfn>ThingDescription</dfn> type</h2>
         <a>TD</a>.
       </p>
       <div>
-        To <dfn>expand a TD</dfn> given |td:ThingDescription|, run the following steps:
```
relu91, line 335: I think this <dfn> element should be kept, is it @zolkis?

index.html
```diff
@@ -316,7 +316,7 @@ <h2>The <dfn>ThingDescription</dfn> type</h2>
           let res = await fetch('https://tds.mythings.biz/sensor11');
           // ... additional checks possible on res.headers
           let td = await res.json();
-          let thing = new ConsumedThing(td);
+          let thing = await WOT.consume(td);
```
zolkis, line 319: Although it is possible to use the constructor, it is more correct to use the factory method, indeed.

index.html
zolkis, line 335: Why to remove the algorithm name?

index.html
```diff
@@ -347,23 +347,24 @@ <h2>The <dfn>ThingDescription</dfn> type</h2>
         Therefore, this API expects the {{ThingDescription}} objects be validated before used as parameters. This specification defines a basic <a>TD</a> validation as follows.
       </p>
       <div>
-        To <dfn>validate a TD</dfn> given |td:ThingDescription|, run the following steps:
```
zolkis, line 350: Same: I don't think algorithm names and params should be removed.

index.html
```diff
@@ -347,23 +347,24 @@ <h2>The <dfn>ThingDescription</dfn> type</h2>
...
             If any of the mandatory properties defined in [[!WOT-TD]] for
             <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#thing">Thing</a>
             that don't have <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#sec-default-values">default definitions</a>
             are missing from |td|, [= exception/throw =] a
             {{"TypeError"}} and abort these steps.
           </li>
-          <li>
-            If <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#json-schema-for-validation">JSON schema validation</a> fails on |td|, [= exception/throw =] a
-            {{"TypeError"}} and abort these steps.
-          </li>
```
relu91, line 366: @egekorkan comment on this change:
> The JSON Schema validation should be before step 2 and that way if something fails the schema validation, it must have values for terms that do not have defaults. The schema validation does not expect a TD with defaults filled in.

relu91: I'd leave as it was. As far as I understand if a property is marked as *required* in JSONSchema even if it has a default value, it must appear in the validating object. However, I'd add a step to make the process a little bit clearer:
1. Check if it is an Object
2. Check if any of the mandatory fields with no default are not defined.
3. Fill mandatory fields with default
4. Do JSONSchema validation

index.html
relu91: I agree, see my review.

index.html
zolkis: Yes.

relu91:
I added a filling default step in the validation algorithm. This is because AFAIK JSONschema validation fails if a required property is not present even if it has a default value. @egekorkan could you check this?

relu91:
It seems there is a relevant discussion about this topic in https://github.com/eclipse/thingweb.node-wot/issues/302

index.html
```diff
@@ -360,6 +360,10 @@ <h2>The <dfn>ThingDescription</dfn> type</h2>
             are missing from |td|, [= exception/throw =] a
             {{"TypeError"}} and abort these steps.
           </li>
+          <li>
```
zolkis, line 363: This section needs to be before the previous step:
1. fill in the default values
2. check for mandatory values.
There might be a case that a mandatory value does have a default value.

egekorkan:
The thing is that the values that have a default value are not required in the official validation schema, i.e. if a property form does not have an `op` value, it will pass validation. Thus, an algorithm that fills the defaults should check that a TD is valid in the first place. In Thingweb Playground, the first schema validation corresponds to the validation with the official schema whereas the second validation that says with defaults correspond to a validation where all the values that have defaults should be filled. The previous example with a property with no `op` fails the second validation.

index.html
```diff
@@ -360,6 +360,10 @@ <h2>The <dfn>ThingDescription</dfn> type</h2>
...
+            Fill any missing mandatory property defined in [[!WOT-TD]] that have a 
+            <a href="https://www.w3.org/TR/2019/CR-wot-thing-description-20190516/#sec-default-values">default definition</a> with its default.
```
zolkis, line 365: Suggestion: usually these complex steps should be more detailed, e.g. using sub-steps with a loop, or an explicit loop, for instance

> For all mandatory properties defined in [[!WOT-TD]] that have default definitions, add the property to |td| if it's missing and set its value to the default value specified in [[!WOT-TD]].

index.html
zolkis: Looks like we cross-commented with @egekorkan here, and they are opposite. Ege, please check if my comment above is valid.

index.html
zolkis: I agree with the sequence proposed above.

index.html
zolkis: I checked the TD spec and it seems there are no cases for mandatory properties that would have default values.
Therefore we can omit the filling up step, and also we can go with Ege's proposal.

index.html
zolkis: Looks like there are no mandatory properties in the TD spec that would have default values, so this text needs updating. We need an initialization policy for the default values of mandatory TD properties. I suggest we create a separate issue about that in the TD repo, then link it here and also in a new issue in this repo. Then we can resolve this and merge this PR.

relu91:
> In Thingweb Playground, the first schema validation corresponds to the validation with the official schema whereas the second validation that says with defaults correspond to a validation where all the values that have defaults should be filled. The previous example with a property with no op fails the second validation.

The thing here is that if it has a default, it probably does not need to be defined.... I agree with @danielpeintner here:
>However I don't think that any implementation needs to add these default values. What would be the idea of having them in the first place if they NEED to get added. My take on this is that IF there is no value the default values kicks in for any processing. The reason to NOT HAVE them is manifold:
>
>* reduced data exchange (TD more compact)
>
>* simpler to see *non* defaults
>
>* avoiding to exchange knowledge that is default anyway

relu91:
Meanwhile, I updated the validation algorithm to have a really basic starting point. I removed the following steps:
> 1. If td is not an object, throw a "TypeError" and abort these steps.
> 2. If any of the mandatory properties defined in [WOT-TD] for Thing that don't have default definitions are missing from td, throw a "TypeError" and abort these steps.

They are both covered by the JSONSchema validation process. The only difference is that JSONSchema is more restrictive on point 2. It fails even if the mandatory property has a default assigned. However, since  `Looks like there are no mandatory properties in the TD spec that would have default values ` I think we can omit it.

PR #235, opened 2020-08-17
feat: update typescript definitions

danielpeintner:


danielpeintner:
@zolkis  @relu91 please have a look. I hope I did not miss anything ;-)

zolkis:
At first sight looks good, but will check more thoroughly tomorrow.

danielpeintner:
> Looks good. However, should we mark read/writeMultiple deprecated?

we could add @deprecated to the method call descriptions but at the moment we are referring to TD 1.0 which does *not* speak about any deprecation... mhh

danielpeintner:
I tend to merge the PR now "as is" to facilitate the start of updating node-wot.

I will publish it on NPM and as `"version": "0.8.0-SNAPSHOT.1"` and we can still update if issues pop up..

danielpeintner:
FYI: see https://www.npmjs.com/package/wot-typescript-definitions

PR #262, opened 2020-09-14
Fix #258: improve expose() algorithm

### LINKED ISSUE #258: Check supported protocols in the validation algorithm
As discussed in today's call, we are missing a validation step for TDs regarding the ability to implement a particular security scheme. Currently, this check is hidden behind the 7th step of the [expose algorithm](https://w3c.github.io/wot-scripting-api/#the-expose-method):
> Set up the WoT Interactions based on introspecting |td| as explained in [WOT-TD] and [WOT-PROTOCOL-BINDINGS]. Make a request to the underlying platform to initialize the Protocol Bindings and then start serving external requests for WoT Interactions (read, write and observe Properties, invoke Actions and manage Event subscriptions), based on the Protocol Bindings.

I think we can make this more explicit and add a separate step where we fail to expose if the underlying platform does not support *one or more* protocol schemes defined in exposing TD. Another possibility is that we fail only if *none* of the provided protocol schemes are supported. Or, according to the latest changes in the TD spec ([here](https://github.com/w3c/wot-thing-description/pull/944)), we have to compute a boolean expression to know the set of security schemes required by the TD and fail if one of them is not supported.  

A final possibility, introduced by @danielpeintner (please correct me if I am wrong), is to treat the whole TD passed as an argument to the ExposedThing constructor as a suggestion and ignore fields and properties that cannot be satisfied. This is what node-wot is doing for `forms`. (See also my comment https://github.com/w3c/wot-scripting-api/issues/238#issuecomment-682060035)

Related to: #238.

zolkis:
This requires more discussion, e.g. how do we check that each interaction has a Form from each protocol mentioned in overall set of Forms. In this PR I left that check to the implementation's discretion.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-scripting-api/pull/262.html" title="Last updated on Sep 14, 2020, 2:25 PM UTC (cf5bff0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/262/c8ae596...zolkis:cf5bff0.html" title="Last updated on Sep 14, 2020, 2:25 PM UTC (cf5bff0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2879,9 +2879,6 @@
             for <a>WoT Interactions</a> (read, write and observe <a>Properties</a>,
             invoke <a>Action</a>s and manage <a>Event</a> subscriptions),
             based on the <a>Protocol Bindings</a>.
-            If for any <a>WoT Interaction</a> there is no |form| specified that
-            is supported by the implementation, reject |promise| with a
-            {{NotSupportedError}} and abort these steps.
             Implementations MAY reject this step for any reason (e.g. if they
             want to enforce further checks and constraints on interaction forms).
           </li>
--- a/index.html
+++ b/index.html
@@ -1696,7 +1696,7 @@
         } catch (e) {
           console.log("TD fetch error: " + e.message);
         };
-        
+
         try {
           // subscribe to property change for “temperature”
           await thing.observeProperty("temperature", async (data) => {
@@ -1722,13 +1722,13 @@
         } catch (e) {
           console.log("Error starting measurement.");
         }
-        
+
         setTimeout(async () => {
           try {
             const temperatureData = await thing.readProperty("temperature")
             const temperature = await temperatureData.value();
             console.log("Temperature: " + temperature);
-        
+
             await thing.unsubscribe("ready");
             console.log("Unsubscribed from the ‘ready’ event.");
           } catch (error) {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2874,7 +2874,16 @@ <h3>The <dfn>EventSubscriptionHandler</dfn> callback</h3>
...
+            Make a request to the underlying platform to initialize the
+            <a>Protocol Bindings</a> and then start serving external requests
+            for <a>WoT Interactions</a> (read, write and observe <a>Properties</a>,
+            invoke <a>Action</a>s and manage <a>Event</a> subscriptions),
+            based on the <a>Protocol Bindings</a>.
+            If for any <a>WoT Interaction</a> there is no |form| specified that
+            is supported by the implementation, reject |promise| with a
+            {{NotSupportedError}} and abort these steps.
+            Implementations MAY reject this step for any reason (e.g. if they
+            want to enforce further checks and constraints on interaction forms).
```
danielpeintner, line 2886: These statements make it difficult to expose a TD that points to interactions *not* hosted by this runtime (maybe even with bindings that are not supported by this WoT runtime)

zolkis: Fixed.

danielpeintner (approved):

PR #265, opened 2020-09-23
Proposal for typescript definition file

### LINKED ISSUE #215: [typescript] Use namespace instead of interface for WoT functions

relu91:
I finally found a solution to have our typescript definition file work both in module and no-module environments. Practically this means that `tsc` will validate both of those code snippets:
```ts
WoT.produce(td).then((thing:WoT.ExposedThing)=>{/*...*/});
```
```ts
import {produce,ExposedThing} from "WoT";
produce(td).then((thing:ExposedThing)=>{/*...*/});
```
As a example the first one might be written inside WoT scripts (which may not support a module system) and the latter for runtimes who want to use node-wot as a library.

Since I had to format the file it is quite difficult to undestand the change. Essentially I changed the declaration of the namespace from `export as namespace WoT;` to:
```ts
declare namespace WoT {
/* declarations */
}
```
Finally I exposed every field in the namespace as a module with the following:
```ts
declare module "WoT" {
    export = WoT;
}
```

I created a PR since I think this is the final solution. 

Fixes #215 .

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -1,300 +1,300 @@
-export as namespace WoT;
-
-/**
- * The WoT interface defines the API entry point exposed as a singleton and contains the API methods.
- */
-export interface WoT {
+declare namespace WoT {
     /**
-     * Starts the discovery process that will provide ConsumedThing 
-     * 
-     * @param filter represents the constraints for discovering Things as key-value pairs
-     */
-    discover(filter?: ThingFilter): ThingDiscovery;
+      * Starts the discovery process that will provide ConsumedThing 
+      * 
+      * @param filter represents the constraints for discovering Things as key-value pairs
+      */
+    export function discover(filter?: ThingFilter): ThingDiscovery;
 
     /**
      * Accepts a ThingDescription and returns a ConsumedThing
      * @param td thing description
      */
-    consume(td: ThingDescription): Promise<ConsumedThing>;
+    export function consume(td: ThingDescription): Promise<ConsumedThing>;
 
     /**
      * Accepts a ThingDescription and returns a ExposedThing
      * 
      * @param td thing description 
      */
-    produce(td: ThingDescription): Promise<ExposedThing>;
+    export function produce(td: ThingDescription): Promise<ExposedThing>;
+
+    /**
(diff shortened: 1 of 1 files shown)


zolkis (approved):

danielpeintner (approved):

danielpeintner:
I published it as "0.8.0-SNAPSHOT.2" on [npm](https://www.npmjs.com/package/wot-typescript-definitions).

Moreover, I also updated [v0.8.x](https://github.com/eclipse/thingweb.node-wot/tree/v0.8.x) branch of node-wot but I did run into issues like

`src/thing-description.ts(17,22): error TS2306: File 'C:/Projects/WoT/thingweb/thingweb.node-wot/packages/td-tools/node_modules/wot-typescript-definitions/index.d.ts' is not a module.`

Currently we use the following which does not work anymore.

`import * as WoT from "wot-typescript-definitions";`

@relu91 Can you let me know how it is assumed to work. Thanks!

Note: The TS definitions still miss the *new* Subscription (I will add/update once the issues are resolved).

PR #283, opened 2020-11-11
Fix #280 and 281: add emitPropertyChange(), remove ConsumedThing inheritance

### LINKED ISSUE #280: Make property change notifications explicit in ExposedThing
Coming from this [comment](https://github.com/eclipse/thingweb.node-wot/issues/333#issuecomment-722431442) from https://github.com/eclipse/thingweb.node-wot/issues/333

zolkis:
add explicit emitPropertyChange() and remove ConsumedThing inheritance from ExposedThing

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-scripting-api/pull/283.html" title="Last updated on Nov 16, 2020, 1:50 PM UTC (2306bca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/283/8bdf0c7...zolkis:2306bca.html" title="Last updated on Nov 16, 2020, 1:50 PM UTC (2306bca)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2019,7 +2019,7 @@
     </p>
     <pre class="idl">
       [SecureContext, Exposed=(Window,Worker)]
-      interface ExposedThing: ConsumedThing {
+      interface ExposedThing {
         ExposedThing setPropertyReadHandler(DOMString name,
                 PropertyReadHandler handler);
         ExposedThing setPropertyWriteHandler(DOMString name,
@@ -2028,6 +2028,7 @@
                 PropertyReadHandler handler);
         ExposedThing setPropertyUnobserveHandler(DOMString name,
                 PropertyReadHandler handler);
+        Promise&lt;undefined&gt; emitPropertyChange(DOMString name);
 
         ExposedThing setActionHandler(DOMString name, ActionHandler action);
 
@@ -2042,6 +2043,8 @@
 
         Promise&lt;undefined&gt; expose();
         Promise&lt;undefined&gt; destroy();
+
+        ThingDescription getThingDescription();
       };
 
       callback PropertyReadHandler = Promise&lt;any&gt;(
@@ -2060,6 +2063,30 @@
 
       callback EventListenerHandler = Promise&lt;InteractionInput&gt;();
     </pre>
+
+    <section data-dfn-for="ExposedThing">
+      <h4>Internal slots for ExposedThing</h4>
+      <p>
+        An {{ExposedThing}} object has the following <a>internal slots</a>:
+      </p>
+      <table class="simple">
+        <thead>
+         <tr>
+          <th>Internal Slot</th>
+          <th>Initial value</th>
+          <th>Description (<em>non-normative</em>)</th>
+         </tr>
+        </thead>
+        <tbody data-link-for="ExposedThing">
+         <tr>
+          <td><dfn>[[\extd]]</dfn></td>
+          <td>`null`</td>
+          <td>The <a>Thing Description</a> of the <a>Thing</a>.</td>
+         </tr>
+        </tbody>
+      </table>
+    </section>
+
     <section>
       <h3>Constructing {{ExposedThing}}</h3>
       <p>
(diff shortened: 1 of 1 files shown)


relu91 (approved):
Approve. Minor comments:
- Could we link the first editor's note also in the other algorithms? it might be handy when the document is read not linearly. 
- Could we use camel case for ExposedThing internal slot `extd`. It might be confused with a shortened version of `extend` wherese it should be ExposedThingDescription. So it might be more intuitive to call it `exTD` or `exTd`.

zolkis:
> Could we use camel case for ExposedThing internal slot extd. It might be confused with a shortened version of extend whereas it should be ExposedThingDescription. So it might be more intuitive to call it exTD or exTd.

Makes sense, done.

> Could we link the first editor's note also in the other algorithms? it might be handy when the document is read not linearly.

The algorithms say "selected by the implementation", which should be clear enough even without the ednote. Repeating the ednote everywhere would be too much clutter, given that it just repeats the last sentence. Perhaps we could remove even the first one?

danielpeintner:
> > Could we use camel case for ExposedThing internal slot extd. It might be confused with a shortened version of extend whereas it should be ExposedThingDescription. So it might be more intuitive to call it exTD or exTd.
> 
> Makes sense, done.

Nit: I am always in favor of spelling out "what" it is since exTD sounds like example TD to me. Why not name it along the lines what it is: `exposedTD`.
Since it is internal only I don't mind much.

index.html
```diff
@@ -2019,7 +2029,7 @@ <h2>The <dfn>ExposedThing</dfn> interface</h2>
     </p>
     <pre class="idl">
       [SecureContext, Exposed=(Window,Worker)]
-      interface ExposedThing: ConsumedThing {
+      interface ExposedThing {
```
danielpeintner, line 2032: I wonder whether it makes sense to create an *abstract* interface Thing

```
interface Thing {
  constructor(ThingDescription td);
  ThingDescription getThingDescription();
}
```

which gets extends by `ConsumedThing `and `ExposedThing`.

I believe doing so avoids duplicating the internal slots in both interfaces.

danielpeintner (approved):
I am generally fine with the PR.

I added some comments that seem beneficial to me.

index.html
```diff
@@ -1118,8 +1118,13 @@ <h3>The <dfn>readProperty()</dfn> method</h3>
           <li>
             If |option|'s |formIndex| is defined, let |form| be the
             <a>Form</a> associated with |formIndex| in |interaction|'s |forms|
-            array, otherwise let |form| be the first <a>Form</a> in
-            |interaction|'s |forms| whose |op| is `readproperty`.
+            array, otherwise let |form| be a <a>Form</a> in
+            |interaction|'s |forms| whose |op| is `readproperty`, selected by
+            the implementation.
+            <p class="ednote">
+              Implementations are allowed to define their algorithms on
```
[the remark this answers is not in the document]
zolkis: The `formIndex` case was already handled at this point, no need to mention it here.
It's getting more and more repeating what is written in the proper step, so I'd probably remove this ednote by now.

index.html
zolkis: They are _separate_ objects with separate internal slots. Internal slots are not inherited. 
On the other hand, this practice is discouraged in Web APIs.

index.html
```diff
@@ -1225,7 +1231,8 @@ <h3>The <dfn>readAllProperties()</dfn> method</h3>
           </li>
           <li>
             If |option|'s |formIndex| is defined, let |form| be the
-            <a>Form</a> associated with |formIndex| in [[\td]]'s |forms| array, otherwise let |form| be the first <a>Form</a> in [[\td]]'s |forms| array whose |op| is `readallproperties`.
+            <a>Form</a> associated with |formIndex| in [[\td]]'s |forms| array, otherwise let |form| be a <a>Form</a> in [[\td]]'s |forms| array whose |op| is `readallproperties`, , selected by
```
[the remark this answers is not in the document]
zolkis: fixed

index.html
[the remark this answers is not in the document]
danielpeintner: Agree, removing sounds fine also!

index.html
danielpeintner: As said, I don't have a strong opinion and it is fine like it is now.

Having said that, I am pretty sure that if it were me to implement it I would probably share the code ;-)

zolkis: ExposedThing implementations are free to do whatever that makes sense, including extending the object.

PR #289, opened 2021-01-21
Add definitions for partialTD

### LINKED ISSUE #287: Partial TD validation
The scripting API [`produce()`](https://w3c.github.io/wot-scripting-api/#the-produce-method) methods allows for passing a TD fragment argument. It is essentially a init dictionary of a TD.

We might want to have a proper validation of this TD fragment argument.
The idea is to use the existing [JSON Schema](https://w3c.github.io/wot-thing-description/#json-schema-for-validation) form the ThingDescription document with the only difference that all `"required": { ... }` keywords are pruned. 

Question: difference between TD model and TD fragment used in scripting.
see [related discussion for TD model](https://github.com/w3c/wot-architecture/issues/574#issuecomment-757925291).

relu91:
This PR introduces the new concept of a PartialTD defined in the architecture PR: https://github.com/w3c/wot-architecture/pull/577.
The changes are heavier than I thought therefore I'd like to receive some feedback about the direction that I am taking. 

As you can see I changed the argument type of the `produce` method to `PartialTD`. This is because a Thing Description is also an instance of a `PartialTD`; in fact, the validation algorithm would accept any valid TD. 

Another open point is how we transform a `PartialTD` to a valid `Thing Description`. I tried to sketch an algorithm in this PR; please have look.

Notice finally, how these changes affect the construction of an `ExposedThing`. Now the constructor has first to transform the `PartialTD` and then extend the obtained TD with the default values.  

Fixes #287

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/289.html" title="Last updated on Feb 17, 2021, 2:48 PM UTC (5b7b6c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/289/ef9c211...relu91:5b7b6c8.html" title="Last updated on Feb 17, 2021, 2:48 PM UTC (5b7b6c8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -484,7 +484,7 @@
         <li>Search for missing required properties in |td| accordingly to
           <a href="https://www.w3.org/TR/2020/WD-wot-thing-description11-20201124/#json-schema-for-validation">TD JSON
             Schema</a>.
-          (TODO: possibly expand this step)
+          <p class="ednote">The editors find this step vague. It will be improved or removed in the next iteration. </p>
         </li>
         <li>For each |missing| property run these sub-steps:
           <ol>
@@ -492,10 +492,13 @@
             <li>If |missing| is <code>@context</code> assign the latest supported Thing Description context URI.</li>
             <li>If |missing| is <code>instance</code> assign the string <code>1.0.0</code>.</li>
             <li>If |missing| is <code>forms</code> generate a list of <a>Forms</a> using the available <a>Protocol Bindings</a> and content types
-              encoders. Then assign the obtained list to <code>forms</code>. (TODO: expand?)</li>
+              encoders. Then assign the obtained list to <code>forms</code>.</li>
             <li>If |missing| is <code>security</code> assign the label of the first supported <a>SecurityScheme</a> in <code>securityDefinitions</code> field. 
             If no <a>SecurityScheme</a> is found generate a <a>NoSecurityScheme</a> called <code>nosec</code> and assing the string <code>nosec</code>
-          to <code>security</code>. (TODO: ask for guidance to Security task force)  </li>
+          to <code>security</code>. 
+          <p class="issue">The discussion about how to properly generate a value for <code>security</code> is still open. 
+            See issue <a href="https://github.com/w3c/wot-scripting-api/issues/299">#299</a> </p>
+          </li>
             <li>If |missing| is <code>href</code> define |formStub| as the partial <a>Form</a> that does not have <code>href</code>. Generate a valid |url:URL| using the first <a>Protocol Binding</a> 
               that satisfy the requirements of |formStub|. Assign |url| to <code>href</code>. If not <a>Protocol Binding</a> can be found remove |formStub| from |td|. </li>
             <li>Add |missing| to |td| with |value| as value</li>
--- a/index.html
+++ b/index.html
@@ -515,7 +515,7 @@
       To <dfn>expand an ExposedThingInit</dfn> given |init:ExposedThingInit| and obtain a valid |td:ThingDescription| as
       a result,
       run the following steps:
-      <ol>
+      <ol class="algorithm">
         <li>Run <a>validate an ExposedThingInit</a> on |init|. If that fails,
           [= exception/throw =] {{SyntaxError}} and abort these steps.</li>
         <li>Initialize and empty object called |td|</li>
@@ -570,7 +570,7 @@
     <section>
       <h3>Validating an ExposedThingInit</h3>
       To <dfn>validate an ExposedThingInit</dfn> given |init:ExposedThingInit|, run the following steps:
-      <ol>
+      <ol  class="algorithm">
         <li>
           Parse <a href="https://www.w3.org/TR/2020/WD-wot-thing-description11-20201124/#json-schema-for-validation">TD JSON
             Schema</a>
@@ -585,8 +585,8 @@
             <li>if |key| |value| is a <code>string</code> then if |value| is equal to one of the elements in |optional| remove |key| from |exposedThingInitSchema|</li>
           </ol>
         </li>
-        <li>Return the result of <a>validating</a> |init| with |exposedThingInitSchema|
-          <p class="ednote"><dfn>Validating</dfn> steps are still under discussion.
+        <li>Return the result of <a>validating an object with JSON Schema</a> given |init| and |exposedThingInitSchema|. 
+          <p class="ednote">The<dfn>validating an object with JSON Schema</dfn> steps are still under discussion.
             Currently this specification reference to the validation process of JSONSchema. Please
             follow this <a href="https://json-schema.org/draft/2019-09/json-schema-validation.html">document</a> 
             when validating |init| with |exposedThingInitSchema|. Notice that the working group is evaluating an alternative formal approach.
--- a/index.html
+++ b/index.html
@@ -585,7 +585,13 @@
             <li>if |key| |value| is a <code>string</code> then if |value| is equal to one of the elements in |optional| remove |key| from |schema|</li>
           </ol>
         </li>
-        <li>Return the result of applying |schema| to |init|</li>
+        <li>Return the result of <a>validating</a> |init| with |schema|
+          <p class="ednote"><dfn>Validating</dfn> steps are still under discussion.
+            Currently this specification reference to the validation process of JSONSchema. Please
+            follow this <a href="https://json-schema.org/draft/2019-09/json-schema-validation.html">document</a> 
+            when validating |init| with |schema|. Notice that the working group is evaluating an alternative formal approach.
+          </p>
+        </li>
       </ol>
     </section>
   </section>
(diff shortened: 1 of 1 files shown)


zolkis:
The first thing I asked myself, from the point of view of an external user not fully familiar with WoT TD, could we use simpler terms?  `PartialTD` looks bad as a type name.

`produce()` needs an init dictionary and the algorithm tells how to use it. Its type is `object`, its semantic is 'partial TD', but in JS we often call init dictionaries with the type+Init suffix. So I think it would be more idiomatic to say

`Promise<ExposedThing> produce(ThingDescriptionInit init)` 

and explain in the text that `init` is a partial TD in WoT Arch terminology.

relu91:
At first, the suggestion seemed reasonable, but now I wonder if having yet another term to specify the same concept would increase the confusion. @danielpeintner what do you think?

zolkis:
Another option could be to continue using the term `ThingDescription`, but mention in the text that the algorithm also accepts a Partial TD (links to definition). Also the internal algorithm names can be kept.

danielpeintner:
Mhh, I tend to agree with Zoltan. We should not overly complicated it.
Sticking with `ThingDescription` and explaining that the input may be a *partial* TD in the case of `expose()` seems like a valid approach.

relu91:
> Another option could be to continue using the term ThingDescription, but mention in the text that the algorithm also accepts a Partial TD (links to definition). Also the internal algorithm names can be kept.

I would avoid this option because `ThingDescription` would be confused with the Thing Description requirements. A PartialTD is not a Thing Description. We clearly say that the `ThingDescription` type :
> Represents a Thing Description (TD) as defined in [WOT-TD]. It is expected to be a parsed JSON object that is validated using JSON Schema validation.

If we stick with that definition (wich works for every other aspect and I don't think we should change it) and leave the produce method signature like it is in the current document, we can't use a PartialTD as argument.  

In particular we cannot state the following:
>Sticking with ThingDescription and explaining that the input may be a partial TD in the case of expose() seems like a valid approach.

It would be like saying that the input type is a Cat but you could use any Animal as input.

zolkis:
> It would be like saying that the input type is a Cat but you could use any Animal as input.

The argument does make logical sense, but I still think `PartialTD` is not a desirable name.

A JS developer will see an object. They pass a generic JS object that satisfies the specified validation algorithm (which may consider partial TDs, full TDs, etc).

```javascript
Promise<ExposedThing> produce(object init);
```

but usually this is not recommended. 
We could also give it a type name that avoids the TD confusions (and explain in prose where it maps): 

```javascript
typedef object ThingInit;
Promise<ExposedThing> produce(ThingInit init);
```

egekorkan:
Well, something that worries me a bit is that I feel that we are being flexible but unclear at the same time. There are issues in node-wot that arise simply due to this confusion:
- https://github.com/eclipse/thingweb.node-wot/issues/375
- https://github.com/eclipse/thingweb.node-wot/issues/364

There can be some other ones that I don't remember but these were the recent ones. So if the discussion is about whether saying partialTD or not, I think that can be changed but saying 

>Another option could be to continue using the term ThingDescription, but mention in the text that the algorithm also accepts a Partial TD (links to definition)

does not fix the problem. I think it should say that it accepts partial TDs and it is not guaranteed that a full TD is taken in its entirety.

relu91:
Totally agree with your comment @egekorkan , that's why I started a dedicated section where to describe how the runtime should treat the PartialTD to generate a valid TD (Implementing a PartialTD). The algorithm is still a WIP but I think we should answer also the questions contained in: 

- https://github.com/eclipse/thingweb.node-wot/issues/375
- https://github.com/eclipse/thingweb.node-wot/issues/364

danielpeintner:
Scripting Call 2021-01-25: replace `PartialTD` type with `ExposedThingInit`.

we might improve algorithmns step-by-step (by means of JSON Schema, ?JSON Schema Patch?)
@ashimura mentions the need of more examples, e.g., how to use partial TDs

danielpeintner:
The general approach looks good to me 👍 

A general comment about the new section headers. We should use
* Use and ExposedThingInit / Validate an ExposedThingInit  OR
* Using and ExposedThingInit / Validating an ExposedThingInit

I wonder how we can make best progress w.r.t. the remaining TODOs. Shall we try to look into some of those to start with?

relu91:
I tried to improve the algorithm for the generation of a TD from an `ExposedThingInit`. The idea is to have a first step where the algorithm removes all the values that couldn't be satisfied by the runtime and then generate the missing required properties. 

Some issues that I have:
- Remove and replace user values feels a little bit intrusive to me. I followed this approach only because it is the _standard de facto_ right now
- Some statements are still a little too vague. We can improve them after discussion.
- I added a final validation step to assure that the generated values are correct. I am not sure to keep it, should we trust runtime generated values? 

I am open to comments/suggestions.

index.html
```diff
@@ -437,13 +444,91 @@ <h2>The <dfn>WOT</dfn> namespace</dfn></h2>
...
-          Let |thing:ExposedThing| be a new {{ExposedThing}} object constructed with |td|.
+          Let |thing:ExposedThing| be a new {{ExposedThing}} object constructed with |init|.
         </li>
         <li>
           Resolve |promise| with |thing|.
         </li>
       </ol>
     </div>
+    <section>
+      <h3>Use an ExposedThingInit</h3>
```
zolkis, line 455: I am thinking if "Expand an ExposedThingInit" would be more accurate.

index.html
```diff
@@ -437,13 +444,91 @@ <h2>The <dfn>WOT</dfn> namespace</dfn></h2>
...
+        <li>
+          For each property |key| in |schema|,
+          <ol>
+            <li>If |key| equals <code>required</code> then remove |key| from |schema|</li>
+          </ol>
+        </li>
+        <li>Return the result of applying |schema| to |init|</li>
+      </ol>
+    
+      TODO: is there any property that we should still require? for example flow property is mandatory for oAuth Security
```
zolkis, line 528: Instead of the TODO, please use a `<p class="issue"> </p>` or `<p class="ednote"> </p>`

zolkis (commented):
Overall good, with minor nits (commented separately).

In general, please use line breaks/wrap around 80 chars, in order to ease review.
Though we applied this quite intermittently in the past. Make a choice :)

danielpeintner:
Scripting Call 2021-02-08
- keep PR open, try to resolve some of TODOs
- use editorial note instead of TODO
- guidance security schemes (ask @mmccool)
- how to make sure user suggestions are fulfilled? 
- improve example (last step)

relu91:
Input from @mmccool about missing `security` field in partial tds:
- Applications should avoid entirely providing `securityDefinitions`
- They may choose a particular definition using the `schema` field (globally or at form level)
- There might be _managment_ APIs to definite the default `SecuirtySchema` among the possible  `securityDefinitions`

I'll open two more issues to keep track of other comments/insights that came up during the security call.

index.html
```diff
@@ -585,7 +585,13 @@ <h3>Validating an ExposedThingInit</h3>
             <li>if |key| |value| is a <code>string</code> then if |value| is equal to one of the elements in |optional| remove |key| from |schema|</li>
           </ol>
         </li>
-        <li>Return the result of applying |schema| to |init|</li>
+        <li>Return the result of <a>validating</a> |init| with |schema|
```
zolkis: I suggest this:

> Return the result of <a>validating an object with JSON Schema</a> given |init| and |schema|

index.html
```diff
@@ -585,7 +585,13 @@ <h3>Validating an ExposedThingInit</h3>
...
+          <p class="ednote"><dfn>Validating</dfn> steps are still under discussion.
+            Currently this specification reference to the validation process of JSONSchema. Please
+            follow this <a href="https://json-schema.org/draft/2019-09/json-schema-validation.html">document</a> 
+            when validating |init| with |schema|. Notice that the working group is evaluating an alternative formal approach.
+          </p>
```
zolkis, line 593: Then also this needs some update:

```
<p class="ednote"> 
  The<dfn>validating an object with JSON Schema</dfn> steps are still under discussion.
  Currently this specification uses the <a href="https://json-schema.org/draft/2019-09/json-schema-validation.html">validation process of JSON Schema</a>. The working group is evaluating an alternative formal approach.
</p>
```

Alternatively, we could already factor out the algo in a subsection, just having an empty body with an ednote, like this:

```
To <dfn>validate an object with JSON Schema</dfn>, given |object| and |schema|, run the following steps:
<ol class="algorithm"
  <li>
     Return the result of <a href="https://json-schema.org/draft/2019-09/json-schema-validation.html">validating</a>
     |init| using |schema|.
     <p class = "ednote">
         ...
     </p>
  </li>
</ol>
```

relu91: Thank you for the input! I kept the editor's note because I was not sure where to create the section. We could still add it later.

zolkis (approved):

danielpeintner:
Scripting Call 2021-02-22
* decided to merge PR
* go with additional smaller PRs (if needed)

Issue #290, opened 2021-01-21
PropertyReadHandler properly defined?


danielpeintner:
The [ExposedThing interface](https://w3c.github.io/wot-scripting-api/#the-exposedthing-interface) defines the `PropertyReadHandler` as follows

```
callback PropertyReadHandler = Promise<any>(
        optional InteractionOptions options = null);
```

while  the `ActionHandler` is defined as follows.

```
callback ActionHandler = Promise<InteractionInput>(
        InteractionOutput params,
        optional InteractionOptions options = null);
```

Shouldn't `PropertyReadHandler` be also in the same fashion? Replacing `any` with `InteractionInput`?

```
callback PropertyReadHandler = Promise<InteractionInput>(
        optional InteractionOptions options = null);
```

relu91:
Yes, it should be fixed.

danielpeintner:
Thanks for the feedback. Let's quickly discuss it in our next call.
If we all agree I will provide a PR...

danielpeintner:
resolved by https://github.com/w3c/wot-scripting-api/pull/291

PR #291, opened 2021-01-25
fix: PropertyReadHandler returning promise with InteractionInput

danielpeintner:
see https://github.com/w3c/wot-scripting-api/issues/290
aligns TS definitions also (no NPM publish yet)

***
<a href="https://pr-preview.s3.amazonaws.com/danielpeintner/wot-scripting-api/pull/291.html" title="Last updated on Jan 25, 2021, 1:26 PM UTC (c7fe51d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/291/6c88d6b...danielpeintner:c7fe51d.html" title="Last updated on Jan 25, 2021, 1:26 PM UTC (c7fe51d)">Diff</a>

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -290,13 +290,13 @@ declare namespace WoT {
         getThingDescription(): ThingDescription;
     }
 
-    export type PropertyReadHandler = (options?: InteractionOptions) => Promise<any>;
+    export type PropertyReadHandler = (options?: InteractionOptions) => Promise<InteractionInput>;
 
-    export type PropertyWriteHandler = (value: InteractionOutput, options?: InteractionOptions) => Promise<any>;
+    export type PropertyWriteHandler = (value: InteractionOutput, options?: InteractionOptions) => Promise<undefined>;
 
     export type ActionHandler = (params: InteractionOutput, options?: InteractionOptions) => Promise<InteractionInput>;
 
-    export type EventSubscriptionHandler = (options?: InteractionOptions) => Promise<void>;
+    export type EventSubscriptionHandler = (options?: InteractionOptions) => Promise<undefined>;
 
     export type EventListenerHandler = () => Promise<InteractionInput>;
 
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -292,11 +292,11 @@ declare namespace WoT {
 
     export type PropertyReadHandler = (options?: InteractionOptions) => Promise<InteractionInput>;
 
-    export type PropertyWriteHandler = (value: InteractionOutput, options?: InteractionOptions) => Promise<undefined>;
+    export type PropertyWriteHandler = (value: InteractionOutput, options?: InteractionOptions) => Promise<void>;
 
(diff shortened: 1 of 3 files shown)


typescript/index.d.ts
```diff
@@ -290,13 +290,13 @@ declare namespace WoT {
...
-    export type PropertyReadHandler = (options?: InteractionOptions) => Promise<any>;
+    export type PropertyReadHandler = (options?: InteractionOptions) => Promise<InteractionInput>;
 
-    export type PropertyWriteHandler = (value: InteractionOutput, options?: InteractionOptions) => Promise<any>;
+    export type PropertyWriteHandler = (value: InteractionOutput, options?: InteractionOptions) => Promise<undefined>;
 
     export type ActionHandler = (params: InteractionOutput, options?: InteractionOptions) => Promise<InteractionInput>;
 
-    export type EventSubscriptionHandler = (options?: InteractionOptions) => Promise<void>;
+    export type EventSubscriptionHandler = (options?: InteractionOptions) => Promise<undefined>;
```
relu91, line 299: Even if Web IDL dropped the `void` type, in TS it still makes sense to use it when describing functions that do not return any value. Just my two cents.

relu91 (approved):
Looks good, just a little comment about the usage of `undefined` as return type.

typescript/index.d.ts
danielpeintner: I totally agree, `void` makes more sense to me.

The reason why I changed it is to be consistent. Moreover, I don't think WebIDL dropped it .. it is just no longer recommended I think, see https://github.com/heycam/webidl/pull/906.
We could still use `Void`.

Let's discuss this in today's call.

zolkis: Let's use `undefined` in Web IDL and `void` in TS.

danielpeintner: Scripting Call 2021-01-25: for Typescript we may use `void`

@relu91 the difference is that TS will warn if someone returns in a code block (additional check for code consistency)

danielpeintner:
Good to go ?

typescript/index.d.ts
```diff
@@ -290,9 +290,9 @@ declare namespace WoT {
         getThingDescription(): ThingDescription;
     }
 
-    export type PropertyReadHandler = (options?: InteractionOptions) => Promise<any>;
+    export type PropertyReadHandler = (options?: InteractionOptions) => Promise<InteractionInput>;
 
-    export type PropertyWriteHandler = (value: InteractionOutput, options?: InteractionOptions) => Promise<any>;
+    export type PropertyWriteHandler = (value: InteractionOutput, options?: InteractionOptions) => Promise<void>;
```
zolkis, line 295: `void` --> `undefined` in Web IDL :)

zolkis (approved):

danielpeintner:
FYI: pushed to NPM also, see https://www.npmjs.com/package/wot-typescript-definitions

PR #297, opened 2021-02-08
make TS definitions for DataSchemaValue more accurate

danielpeintner:


typescript/index.d.ts
```diff
@@ -77,7 +77,7 @@ declare namespace WoT {
     export type ThingDescription = { [key: string]: any; };
 
 
-    export type DataSchemaValue = any;
+    export type DataSchemaValue = (string | number | any[] | object | null);
```
danielpeintner, line 80: I am unsure about the array.

We could change `any[]` to `Array<Type>` where `Type` would be `any` or the list of types again? Not sure what makes more sense.

I am more in favor to use the `plain` array construct vs the class style...

relu91: what's wrong with `DataSchemaValue[]`? ts cannot handle recursive declarations?

relu91: https://www.typescriptlang.org/play?#code/C4TwDgpgBAIghsOBlAxgCwgWzgNTgGwFdoBeKACgGdgAnASwDsBzKAHygcMwCMIa3YCZOiy4CxANoBdAQHtuAKwgpgAzvnwBKANwAoXfgiqAbuIgAuQYlQZseIqSgSARBg2znU7UA

it seems to work

danielpeintner: Good point. I missed the obvious :-(

danielpeintner:
@zolkis new order update
= (null | boolean | number | string | object | DataSchemaValue[]);

danielpeintner:
> @zolkis new order update
> = (null | boolean | number | string | object | DataSchemaValue[]);

Updated PR, please approve. Thanks!

relu91 (approved):

PR #305, opened 2021-02-22
Factor out maplike interfaces property reads and writes

zolkis:
It also did some white space corrections automatically.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-scripting-api/pull/305.html" title="Last updated on Mar 1, 2021, 9:20 AM UTC (e3529c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/305/5d70e76...zolkis:e3529c9.html" title="Last updated on Mar 1, 2021, 9:20 AM UTC (e3529c9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1851,14 +1851,20 @@
     <section>
       <h3>The <dfn>PropertyReadMap</dfn> type</h3>
       <p>
-        Represents a map of <a>Property</a> names as strings to a value that the <a>Property</a> can take. It is used as a property bag for interactions that involve multiple <a>Properties</a> at once.
+        Represents a map of <a>Property</a> names to an <a>InteractionOutput</a>
+        object that represents the value the <a>Property</a> can take. It is
+        used as a property bag for interactions that involve multiple
+        <a>Properties</a> at once.
       </p>
     </section>
 
     <section>
       <h3>The <dfn>PropertyWriteMap</dfn> type</h3>
       <p>
-        Represents a map of <a>Property</a> names as strings to a value that the <a>Property</a> can take. It is used as a property bag for interactions that involve multiple <a>Properties</a> at once.
+        Represents a map of <a>Property</a> names to an <a>InteractionInput</a>
+        that represents the value the <a>Property</a> can take. It is
+        used as a property bag for interactions that involve multiple
+        <a>Properties</a> at once.
       </p>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -488,12 +488,12 @@
       };
     </pre>
     <div>
-      Belongs to the <a>WoT Producer</a> conformance class. Expects a |init:ExposedThingInit| argument and returns a {{Promise}} 
-      that resolves with an {{ExposedThing}} object that extends {{ConsumedThing}} with a server interface, 
+      Belongs to the <a>WoT Producer</a> conformance class. Expects a |init:ExposedThingInit| argument and returns a {{Promise}}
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1174,7 +1177,15 @@ <h2>The <dfn>ConsumedThing</dfn> interface</h2>
...
-      typedef object PropertyMap;
+      [SecureContext, Exposed=(Window,Worker)]
+      interface PropertyReadMap {
+        readonly maplike&lt;DOMString, InteractionOutput&gt;;
+      };
+
+      [SecureContext, Exposed=(Window,Worker)]
+      interface PropertyWriteMap {
+        readonly maplike&lt;DOMString, InteractionInput&gt;;
+      };
```
danielpeintner, line 1188: The WebIDL syntax does not seem very intuitive to me. I would have expected something like

`interface PropertyReadMap extends Map<DOMString, InteractionOutput>`.

Anyhow, `WebIDL maplikes` are to be described like this in the body.

Even though I raised the issue I am not sure myself whether if it is worth using maplikes ...

zolkis: Web IDL syntax. https://heycam.github.io/webidl/#idl-maplike

zolkis: This is the way to express the idea in Web IDL, still better than using `object`.

zolkis:
A nice thing about `maplike` is that it's well defined, for instance iterator/forEach is [included](https://heycam.github.io/webidl/#es-iterators).

index.html
```diff
@@ -1838,15 +1849,18 @@ <h3>
     </section>
 
     <section>
-      <h3>The <dfn>PropertyMap</dfn> type</h3>
+      <h3>The <dfn>PropertyReadMap</dfn> type</h3>
       <p>
         Represents a map of <a>Property</a> names as strings to a value that the <a>Property</a> can take. It is used as a property bag for interactions that involve multiple <a>Properties</a> at once.
```
relu91, line 1854: I am not sure that _value_ is the right term here. With the current API model, we return a map of strings and InteractionOutput which is a _gateway_ to the real value. Should we rephrase to something like this:

" Represents a map of Property names as strings to an InteractionOutput that _can be used to retrieve_/_contains_ the value of Property"?

index.html
```diff
@@ -1838,15 +1849,18 @@ <h3>
...
-         It could be defined in Web IDL (as well as {{ThingDescription}}) as a <a href="https://heycam.github.io/webidl/#idl-maplike">maplike</a> interface from <i>string</i> to <i>any</i>.
-      </p>
     </section>
 
+    <section>
+      <h3>The <dfn>PropertyWriteMap</dfn> type</h3>
+      <p>
+        Represents a map of <a>Property</a> names as strings to a value that the <a>Property</a> can take. It is used as a property bag for interactions that involve multiple <a>Properties</a> at once.
+      </p>
+    </section>
```
relu91, line 1863: See my other comment.

relu91 (approved):
The PR it's fine. Some minor discussion points to be addressed.

relu91 (approved):

danielpeintner:
Note: IF we merge this PR we should update TS definitions also (can be done in a subsequent path)

relu91:
> Note: IF we merge this PR we should update TS definitions also (can be done in a subsequent path)

The PR above should covert the issue ✔️

danielpeintner:
Scripting Call 2021-03-01: Syntactic changes seems OK.

PR #307, opened 2021-03-01
fix 302

relu91:
Let actions return undefined in typescript.


danielpeintner (approved):

danielpeintner:
Scripting Call 2021-03-01
* TS requires undefined to be defined, see https://www.typescriptlang.org/play shows errors for the following snippet
```TS
function test():Object {
    return undefined;
}
```

Agreement to merge.

PR #308, opened 2021-03-01
Introduce PropertyReadMap and PropertyWriteMap in TS

relu91:
Synchronize Typescript types with the latest changes in #305

### DIFF
--- a/typescript/index.d.ts
+++ b/typescript/index.d.ts
@@ -180,8 +180,8 @@ declare namespace WoT {
         data?: any;
     }
 
-    export type PropertyReadMap = Map<string,InteractionOutput>
-    export type PropertyWriteMap = Map<string,InteractionInput>
+    export type PropertyReadMap = Map<string, InteractionOutput>;
+    export type PropertyWriteMap = Map<string, InteractionInput>;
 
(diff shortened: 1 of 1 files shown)


typescript/index.d.ts
```diff
@@ -180,7 +180,8 @@ declare namespace WoT {
         data?: any;
     }
 
-    export type PropertyMap = object | { [key: string]: any; };
+    export type PropertyReadMap = Map<string,InteractionOutput>
+    export type PropertyWriteMap = Map<string,InteractionInput>
```
danielpeintner, line 184: nits:
* spaces between arguments -> `Map<string, InteractionOutput>`
* `;` in the end to be consistent with the rest?

danielpeintner (approved):
maybe doing first some consistency updates

danielpeintner:
Scripting Call 2021-03-01:
* content-wise the changes look ok
* @relu91 will update the minor formatting and we can merge

typescript/index.d.ts
relu91: Should be solved now

danielpeintner (approved):

Issue #222, opened 2020-07-03
Possible alternative design of Discovering API


relu91:
Currently, the Discovering API uses the `ThingFilter`'s `method` field to select which discovery approach to use. Therefore, we are using a *factory* style to select different strategies using an Enumerative value. This is common in the `OOP` world but I rather use a more *functional* style to solve this problem. The fact is that `method` field is hinting us that the discovery process could be carried out in different ways, which means that we could model the API using different functions. 

Consequently, my proposal is to remove the `method` field from `ThingFilter` and design a set of functions that represents the different `method` enumerative values. In concrete, I'd define a subnamespace of WoT called `discovery` which contains the `directory`, `local`, `mulitCast`, `direct` (?), and `any` functions. I'll describe the interface formally using Typescript since I am not so used to WebIDL:

```ts
declare namespace WOT {
    namespace discovery {
        function directory(url:[string|ThingDescription],query?:string):ThingDiscovery ;
        function local():ThingDiscovery ;
        function multiCast():ThingDiscovery ;
        function any():ThingDiscovery ;
        function direct(url:string,form?:Form):Promise<ThingDescription>;
    }
}
```

Note that function signatures are just examples, they are not the topic of this issue. 
However, it can be seen how this solution adds flexibility to the inputs/outputs of the different discovery processes. We are not limited to use the `ThingFilter` object containing all the possible inputs of the algorithms. We can now fine-tune the signature to better convey the anatomy of the discovery method. For example, see the (proposal) signature of the `direct` algorithm, where the output is not a ThingDiscovery but just the ThingDescription requested. Or see the `directory` function where we define as input also a ThingDescription of the directory service (if directories will have a TD).

Practically, devs will be able to call the discovery in the following way:
```ts
const TD = await WOT.discovery.direct("http://somehost.some");
const thing = await WOT.consume(TD)
``` 

## pro
- algorithm input/output fine-tuning
- Easier to describe each algorithm. we do not need a big `switch` and explain each case, but we can use different subsections. 
- More expressive. we model the process as a real function, not as an Enum value.
- More extensible. it is easier to change/add discovery algorithms (i.e. adding a new function or change its signature)
- ...?

## cons
- Model `any` function might be hard 
- WebIDL does not seem to support subnamespaces 
- Increase the number of functions in the API (i.e. increase complexity)
- ... ?

#### see also
[...]

PR #316, opened 2021-04-16
Simplify discovery methods. Update examples.

### LINKED ISSUE #311: Simplify discovery methods
Right now, supported [DiscoveryMethod](https://www.w3.org/TR/wot-scripting-api/#dom-discoverymethod) strings are `any`, `directory`, `local`, `multicast`. In earlier drafts we also had `direct`, and #222 considers that use case as well.

I'd like to propose removing `local` and `multicast`, and re-add `direct`, with `any` remaining redundant (removed as well).
So in the end we'd have support for `direct` and `directory` methods.

We had a call with @relu91 about this and we agreed about the following.

Reasons to remove `local`:
- It would be very complex to define what exactly it would mean. It could mean sensors/actuators available on the same HW or SW or WoT runtime. Examples: local accelerometer, various thermometers, gyroscope, proximity, ambient light sensor etc, in general covered by the [Generic Sensor API suite](https://www.w3.org/TR/generic-sensor/). Or NFC adapter, or things locally connected via Bluetooth, etc.
- It would be complex to implement. How a WoT implementation would find all these and how does it expose them as Things?
- This functionality should better be implemented in SW package(s) that need to implement a subset of the locally available sensors and also add Thing Descriptions and (or expose via) a local Thing Directory. 
- The Scripting app, and implementation could then find and integrate a local Thing Directory given its URL.

Reasons to remove `multicast`:
- There is some confusion what does it exactly cover. Multicast discovery is usually used in phase 1 discovery (provisioning time), but in some protocol suites also in operational mode. It is unclear what an implementation should do when receiving `multicast` from an application. Right now the spec is too vague about this: "discovering Things in the device's network by using a supported multicast protocol." It sounds like "discover anything, but with a multicast method". Which leads to the next argument:
- So far we don't have a strong use case for supporting it. Therefore we can test to remove it and re-add later when there is a use case for it.

Reasons to re-add `direct`:
- We have a use case for it in many protocol suites. If an app provides a URL, that is pointing to a Thing Directory, it might provide different things in `direct` mode (e.g. the TD of the directory itself) vs all the TDs contained in the directory (if `directory` was provided as method).
- Provides an more convenient WoT-specific way to fetch a TD from a URL (as opposed to using the Fetch API).

Counter-arguments to re-add `direct`:
- Since discovery in Scripting currently returns TDs, it would be the same as fetching a TD from a URL. Although in the simple cases it would be more convenient for apps to use discovery vs Fetch.
- Using discovery for a "convenient fetch" sounds a bit awkward.

Reasons to remove `any`:
- It is not needed any more. The app has to know if it wants a TD from a URL, or if wants to get TDs from a directory. The directory method could be the default, as the direct method is more specific and it's better visible in the code explicitly.

Please chime in with opinions. I can provide a PR after the discussion settled.

zolkis:
Fixes #311 

Now that discovery methods are simplified, `url` could become a first-class mandatory parameter to `discover()`.
However, that would limit future extension to multicast discovery, if ever comes.
So the current design is still the most generic and robust.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-scripting-api/pull/316.html" title="Last updated on Apr 26, 2021, 7:52 AM UTC (382ba63)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/316/e3529c9...zolkis:382ba63.html" title="Last updated on Apr 26, 2021, 7:52 AM UTC (382ba63)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3570,15 +3570,13 @@
         Represents the discovery type to be used:
       </p>
       <ul>
-        <li><dfn>"any"</dfn> does not provide any restriction</li>
         <li>
-          <dfn>"local"</dfn> for discovering <a>Thing</a>s defined in the same device or connected to the device by wired or wireless means.
+          <dfn>"direct"</dfn> for fetching the <a>Thing Description</a> of a
+          <a>Thing</a> specified by <a>ThingFilter</a>'s `url`.
         </li>
         <li>
-          <dfn>"directory"</dfn> for discovery based on a service provided by a <a>Thing Directory</a>.
-        </li>
-        <li>
-          <dfn>"multicast"</dfn> for discovering <a>Thing</a>s in the device's network by using a supported multicast protocol.
+          <dfn>"directory"</dfn> for discovery based on a service provided by a
+          <a>Thing Directory</a> specified by <a>ThingFilter</a>'s `url`.
         </li>
       </ul>
     </section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3743,21 +3743,19 @@ <h3>Discovery Examples</h3>
       <p>
         The following example finds {{ThingDescription}} objects of <a>Thing</a>s that are exposed by local hardware, regardless how many instances of <a>WoT Runtime</a> it is running. Note that the discovery can end (become inactive) before the internal <a>discovery results</a> queue is emptied, so we need to continue reading {{ThingDescription}} objects until done. This is typical with local and directory type discoveries.
       </p>
-      <pre class="example" title="Discover Things exposed by local hardware">
-        let discovery = new ThingDiscovery({ method: "local" });
+      <pre class="example" title="Fetch the Thing Description of a Thing">
```
relu91, line 3746: Is the explainer text above of this example updated? It refers to local discovery with multiple WoT Runtime a concept that I think we are still missing to formalize.

relu91 (approved):
The PR looks good. I would approve it for merging cause I support the reasoning behind this change (described in #311). 

However, as a subsequent step, I'd like to reevaluate the design described in #222. In particular, it feels a little bit over-complicated to use this current design for the direct method. You have to call the next method of the `ThingDiscovery` when you already know that it will be just one single TD. Besides we have all the pros listed in #222. Now that `any` method is not needed anymore #222 is even more appealing. 

I can provide a PR after we merge this one 👍🏻

index.html
```diff
@@ -3570,15 +3570,14 @@ <h3>The <dfn>DiscoveryMethod</dfn> enumeration</h3>
         Represents the discovery type to be used:
       </p>
       <ul>
-        <li><dfn>"any"</dfn> does not provide any restriction</li>
         <li>
-          <dfn>"local"</dfn> for discovering <a>Thing</a>s defined in the same device or connected to the device by wired or wireless means.
+          <dfn>"direct"</dfn> for fetching the <a>Thing Description</a> of a
+          <a>Thing</a> specified by <a>ThingFilter</a>'s `url`.
+          .
```
danielpeintner, line 3576: minor. 2 dots

index.html
```diff
@@ -3590,24 +3589,31 @@ <h3>The <dfn>ThingFilter</dfn> dictionary</h3>
...
           object? fragment;
+          <!-- DOMString? query; -->
         };
       </pre>
-      <p>
-        The <dfn>method</dfn> property represents the discovery type that should be used in the discovery process. The possible values are defined by the <a href="#the-discoverymethod-enumeration">DiscoveryMethod</a> enumeration that MAY be extended by string values defined by solutions (with no guarantee of interoperability).
+      <p class = "ednote">
+        The `DOMString query` property was temporarily removed from
+        <a>ThingFilter</a>, until it is standardized in the WoT Discovery
+        taskforce.
```
danielpeintner, line 3601: Minor: split to "task force"?

danielpeintner (approved):
LGTM.

I added 2 minor comments that you can take a look at at.

index.html
zolkis: Amended, thanks.

index.html
zolkis: Amended, thanks.

zolkis:
I amended the last commit twice, to correct both minor errors. Merging.

PR #318, opened 2021-05-06
Introduce thing description type from json schema

relu91:
This PR relates to the proposal in https://github.com/w3c/wot-thing-description/issues/1097. As a first step, I manually imported the schema and set up the scripts for translating the schema to a d.ts file. So far so good, the only issue is that the tile of the schema is a bad type name for a TS interface, therefore I manually substitute it using `sed`.  Possibly I'll open another issue on the TD repository to propose a better name and versioning procedure. 

About automatic pulling, my plan is to branch from this PR and trying to set up a GitHub action on my fork.


zolkis:
Thanks, looks good! 
@danielpeintner to approve.

danielpeintner (commented):
some questions
* `package-lock.json` caused some troubles on node-wot side. Do we really want to add this to the repo?
* w.r.t. `td-json-schema-validation.json` does it make sense to point to the original location instead of duplicating it? see https://raw.githubusercontent.com/w3c/wot-thing-description/main/validation/td-json-schema-validation.json

typescript/index.d.ts
```diff
@@ -74,7 +75,7 @@ declare namespace WoT {
      * WoT provides a unified representation for data exchange between Things, standardized in the Wot Things Description specification.
      * In this version of the API, Thing Descriptions is expected to be a parsed JSON object.
      */
-    export type ThingDescription = { [key: string]: any; };
+    export type ThingDescription = ThingDescriptionSchema;
```
danielpeintner, line 78: Note: We need to test but I recall some issues in node-wot if we change it object or so.
Hence we should check that before hands...

danielpeintner:
Scripting Call 2021-05-10:
* merge to test GH actions
* tackle comments later

PR #319, opened 2021-05-06
Add a github action to automatically sync the json schema

relu91:
This PR is based on #318 and its main goal is to automate the process of pulling the json schema from the Thing Description repository. I tested the action on my fork and it works pretty well. What I wasn't able to test was the ability to create the issue, it seems that I cannot tag @danielpeintner on my repository. 

The action is configured to be triggered automatically at 00:00 utc every day. But it has also a convenient button to manually run it (in case we need to test it or we are in hurry to see the changes in our repository 😄 ).


.github/workflows/sync.yml
```diff
@@ -0,0 +1,52 @@
...
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+      - name: Download schema
+        run: | 
+          # Retrive json schema from td repository
```
zolkis, line 19: Retrieve

zolkis:
Good choice anyway, we could also reference these from the spec.

.github/workflows/sync.yml
relu91: Thanks! fixed

danielpeintner:
Scripting Call 2021-05-10:
* merge to test GH actions
* tackle comments later

PR #324, opened 2021-06-04
Fix sync action 

relu91:
Just noticed that the GitHub action did not update the ts file generated but only the schema. There is still an issue with the name of the schema that needs to manually (using bash) replaced. I'll open a PR on the TD repo to use a version field instead of putting it in the title.


danielpeintner (approved):
LGTM

PR #325, opened 2021-06-04
Allign types with the current specification

relu91:
This change should fix problems when trying to use the new types in node-wot or scripts. Furthermore, it aligns the produce method with the current spec. Now it accepts a PartialThingDescription interface defined as a ThingDescription but with everything optional.


danielpeintner (approved):
Question about naming... shall we use the term from the scripting document? 
"PartialThingDescription" -> "ExposedThingInit"  ?

danielpeintner:
let's merge and push to NPM to facilitate testing ...

PR #329, opened 2021-07-26
feat: move thing-description types to a dedicated package

### LINKED ISSUE #327: Publish Thing Description schema and Types in another npm package
As observed in https://github.com/WebThingsIO/gateway-addon-ipc-schema/issues/19#issuecomment-877100054, one problem of the current publication strategy is that it is not clear which version of the Thing Description Schema is bundled in `wot-typescript-definitions package. In addition, projects that just want to use TD types for validation and loading need to carry over also the definitions for the scripting API. 

My suggestion is to split the current package in two:
- wot-api-types (or keep the current naming)
- wot-thing-description-types

Opinions?

relu91:
This PR moves Thing Description types definitions to a dedicated package as discussed in #327. Some remarks:
- The solution uses [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces). As my experience with lerna workspaces was not so pleasant I was a little bit doubtful to use this feature at first. Then I realized that it should be optimal since our `wot-typescript-definitions` package depends on `wot-thing-description-types`. 
- I updated the action to follow the new folder structure
- As we are now using `workspaces` to publish the packages we need to use `npm publish --workspaces` in `typescript` folder
- I'd follow this new procedure when publishing:
  1. Update version in package json
  2. Use git tag to mark the history (e.g. `git tag thing-description@1.1.0-DATE`) 
  3. Publish the package to npm

Fix #327


typescript/package.json
```diff
@@ -1,29 +1,25 @@
 {
-  "name": "wot-typescript-definitions",
-  "version": "0.8.0-SNAPSHOT.16",
-  "description": "TypeScript definitions for the W3C WoT Scripting API",
-  "author": "W3C Web of Things Working Group",
-  "license": "W3C-20150513",
-  "repository": "https://github.com/w3c/wot-scripting-api/tree/master/typescript",
+  "name": "wot-typescript",
+  "version": "1.0.0",
```
danielpeintner, line 3: What are the implications of this version? 
I assume we need to update every time to be able to publish on NPM?

typescript/scripting-api/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "wot-typescript-definitions",
```
danielpeintner, line 2: I wonder whether the folder name (currently `scripting-api` should be aligned with the name)

typescript/thing-description/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "wot-thing-description-types",
```
danielpeintner, line 2: same as above

I wonder whether the folder name (currently `thing-description` should be aligned with the name)

typescript/thing-description/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "version": "1.1.0-26-July-2021",
```
danielpeintner, line 3: do we expect that this kind of versioning breaks somehow NPM since it does not follow their rules, see https://docs.npmjs.com/about-semantic-versioning

Having said that, adding SNAPSHOT doesn't seem to be problematic so it should not cause issues here.. I hope.

typescript/package.json
danielpeintner: Hence, I suggest to change to `0.0.1`

typescript/thing-description/package.json
danielpeintner: Consistency ?
`wot-typescript-definitions`     (exists already)
vs.
`wot-typescript-thing-description`
vs.
`wot-thing-description-types`

Comments?

typescript/scripting-api/package.json
danielpeintner: @relu91 add "private" to package.json so that it cannot be published

danielpeintner:
@zolkis feedback is welcome

zolkis:
Looks good - actually it didn't need my stamp :)

typescript/package.json
danielpeintner: @relu91 OK, also mark as private

typescript/scripting-api/package.json
danielpeintner: Scripting Call 2021-08-09
decided to keep "scripting-api" folder name

typescript/thing-description/package.json
danielpeintner: Scripting Call 2021-08-09
decided to keep "thing-description" folder name

typescript/thing-description/package.json
danielpeintner: Scripting Call 2021-08-09
* keep proposed  version scheme
* check-out whether it causes issues

danielpeintner:
Scripting Call 2021-08-09
* merge after conflicts are resolved

typescript/package.json
relu91: Solved!

typescript/scripting-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "wot-typescript-definitions",
-  "version": "0.8.0-SNAPSHOT.16",
+  "version": "0.8.0-SNAPSHOT.17",
```
danielpeintner, line 3: should be "0.8.0-SNAPSHOT.18"

0.8.0-SNAPSHOT.17 is published already, see https://www.npmjs.com/package/wot-typescript-definitions

relu91: Thank you, is now 18 :)

PR #331, opened 2021-08-11
docs: add readme about TS files and process

danielpeintner:


danielpeintner:
Scripting API Call 2021-08-23
Group decided to merge

PR #332, opened 2021-08-25
fix: links *new* typescript folder

danielpeintner:


danielpeintner:
Scripting API Call 2021-08-30 -> approved

Issue #346, opened 2021-10-18
Clarify multiple requests for subscription/observation


relu91:
With the latest changes, we introduced a new interface for handling `subscriptions`/`observe changing` on a `ConsumerThing`. However, for the implementer is not clear how the runtime should act when an application is issuing multiple subscriptions/obersevation for the same affordance. 

In principle thanks to the new control objection `Subscription` it's possible to handle multiple functions registered as listeners for an affordance. However, this possibility is never mentioned (or at least I couldn't find it) in the document leaving open to interpretation how the process should look like. Furthermore, the document explicitly says that when the `stop` method is called on a `Subscription` this request should be immediately be sent to the `protocol` layer. As I read this, a vague hint to the fact that in practice you can have just one single Subscription object, because as soon as you stop it you have to cancel the subscription at the protocol level (e.g., closing the connection, unsubscribe request etc. ). 

Actions points:
- I remember that we discussed the possibility to have multiple Subscriptions active per affordance, but we discarded it because it would create a more complex API. Should we reconsider this?
- Discuss if this is an implementation detail and should be explicitly left open
- Either way, the document needs few lines discussing the alternatives and, eventually, why we choose one or another.

relu91:
> Question: Is this really a question _we_ as the Scripting API task-force should answer or should that be defined in the TD spec?

I think is up to us, since is something closely related to how an application would use the subscription/observation APIs. 



> We need to add steps to algorithms to clarify this. Option 1
> 
> * subscribing with the same function will do nothing;
> * subscribing with a different function will create a new subscription
> 
> Option 2
> 
> * subscribing with the same function will do nothing
> * subscribing with a different function just changes the subscription callback.
> 
> I think Option 1 is manageable and cleaner. We need to update the unsubscribe steps as well, to map the API behavior to protocol. Some bindings might allow multiple subscriptions with different client side id, some will just allow one kind of subscription since the notifications are the same. Even in the latter case the API might want to support multiple callback functions (one per client-side subscription).

What about Option 3 instead of Option 2?:
- subscribing with the same function will do nothing;
- subscribing with another function will throw an exception 

I think it is safer than silently changing subscription callbacks 🤔 

[...]

PR #356, opened 2021-12-02
Allow only one subscription per affordance

relu91:
This PR adds an additional step to `observeProperty` and `subscribeEvent` algorithms of the `ConsumeThing`. It forces the runtime implementation to check whether the affordance has already an active Subscription, and if it is the case the runtime must throw a `NotAllowedError`. This should make clear to implementers that it is not possible to subscribe multime times to the same affordance, but I also added an editor note to make it cristal clear. 

The PR is currently in draft mode cause I wanted to use [Set](https://tc39.es/ecma262/#sec-set-objects) as type of the list of active Subscriptions/Observations but ECMA specification is not included among our `xref` and it conflicts with `web-platform`. 

*Note*: It covers the points raised in #346.

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/356.html" title="Last updated on Dec 13, 2021, 1:08 PM UTC (667e9c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/356/ec3b3fc...relu91:667e9c9.html" title="Last updated on Dec 13, 2021, 1:08 PM UTC (667e9c9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1677,8 +1677,16 @@
         Takes as arguments |propertyName:string|, |listener:InteractionListener| and
         optionally |onerror:ErrorListener| and |options:InteractionOptions|.
         It returns a {{Promise}} that resolves on success and rejects on failure.
+        <p class="ednote">
+          This algorithm allow for only one active {{Subscription}} per <a>Property</a>. If a new
+          {{Subscription}} is made while an existing {{Subscription}} is active the runtime
+          will throw an {{NotAllowedError}}.
+        </p>
         The method MUST run the following steps:
         <ol>
+          <li>
+            Let |thing| be the reference of this {{ConsumedThing}} object.
+          </li>
           <li>
             Return a {{Promise}} |promise:Promise| and execute the next steps [=in parallel=].
           </li>
@@ -1694,6 +1702,10 @@
             with a {{TypeError}} and abort these steps.
           </li>
           <li>
+            if |thing| [[\activeObservations]] contains |propertyName|, reject |promise| with
+            a {{NotAllowedError}} and abort these steps.
+          </li>
+          <li>
             Let |subscription| be a new {{Subscription}} object with its <a>internal slots</a>
             set as follows:
             <ul>
@@ -1706,6 +1718,9 @@
               <li>
                 Let |subscription|'s [[\interaction]] be the value of [[\td]]'s
                 |properties|'s |propertyName|.
+              </li>
+              <li>
+                Let |subscription|'s [[\thing]] be the value of |thing|.
               </li>
               <li>
                 Let |subscription|'s [[\form]] be the <a>Form</a> associated with
@@ -1734,7 +1749,7 @@
             the <a>Protocol Bindings</a> and abort these steps.
           </li>
           <li>
-            Otherwise resolve |promise|.
+            Otherwise add |propertyName| to |thing|'s' [[\activeObservations]] set and resolve |promise|.
           </li>
           <li>
             Whenever the underlying platform detects a notification for this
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1213,6 +1213,16 @@ <h4>Internal slots for ConsumedThing</h4>
           <td>`null`</td>
           <td>The <a>Thing Description</a> of the {{ConsumedThing}}.</td>
          </tr>
+        <tr>
+          <td><dfn>[[\activeSubscriptions]]</dfn></td>
+          <td>`{}`</td>
```
zolkis, line 1218: It was a good idea to have an internal slot as a set, as it is usable with all policies (replace existing, reject if existing, add, ...).

index.html
```diff
@@ -1739,8 +1764,16 @@ <h3>The <dfn>subscribeEvent()</dfn> method</h3>
         Takes as arguments |eventName:string|, |listener:WoTListener| and
         optionally  |onerror:ErrorListener| and |options:InteractionOptions|.
         It returns a {{Promise}} to signal success or failure.
+        <p class="ednote">
+          This algorithm allow for only one active {{Subscription}} per <a>Event</a>. If a new
```
zolkis, line 1768: Should we repeat this every time, or just say once for the whole API?

zolkis (approved):
Good work, thanks!

index.html
relu91: Good point. I think another good place to state this is in the `Subscription` section but I don't know if we want to keep this simple to possibly change it in the future.

danielpeintner (commented):
I am wondering why this PR includes changes to 
 typescript/thing-description/schema/td-json-schema-validation.json
and 
 typescript/thing-description/thing-description.d.ts 

The changes should be already there... sync with master ?

index.html
zolkis: We can also omit the ednote, since the algorithm is clear enough.

relu91:
Cleaned up commit history and added a fix for #358.

index.html
```diff
@@ -1310,6 +1310,16 @@ <h4>Internal slots for ConsumedThing</h4>
...
          </tr>
+        <tr>
+          <td><dfn>[[\activeSubscriptions]]</dfn></td>
+          <td>`{}`</td>
+          <td>A {{Set}} containing the names of active {{Subscription}}'s for an <a>Event</a></td>
+        </tr>
+        <tr>
+          <td><dfn>[[\activeObservations]]</dfn></td>
+          <td>`{}`</td>
+          <td>A {{Set}} containing the names of active {{Subscription}}'s for an <a>Property</a></td>
```
danielpeintner, line 1321: I guess it should read 
a set of active Observations but then internal linking is no longer  working.... maybe we can refer to it by id

index.html
```diff
@@ -1677,8 +1687,16 @@ <h3>The <dfn>observeProperty()</dfn> method</h3>
         Takes as arguments |propertyName:string|, |listener:InteractionListener| and
         optionally |onerror:ErrorListener| and |options:InteractionOptions|.
         It returns a {{Promise}} that resolves on success and rejects on failure.
+        <p class="ednote">
+          This algorithm allow for only one active {{Subscription}} per <a>Property</a>. If a new
```
danielpeintner, line 1691: typo?: this algorithm allowS ...

index.html
danielpeintner: typo?: this algorithm allowS ...

danielpeintner:
Scripting Call 2021-12-13
@relu91 fixes the 2 editorial issues and than we can merge

relu91:
@danielpeintner it should be ready to go :)

danielpeintner (approved):

PR #372, opened 2022-01-27
Generate types for unreachable definitions

relu91:
https://github.com/w3c/wot-thing-description/pull/1319 introduced a schema definition for `Form`. In principle, this new definition was added to let typescript users have a type for a generic form. Sadly, [json-schema-to-typescript](https://github.com/bcherny/json-schema-to-typescript) does not automatically do it but you have to enable it. This PR enables the generation of unreachable definitions. As you can see on the diff enabling such option causes also the generation of a long list of unwanted comments. I tried to get rid of them without any success. I'm waiting for [an answer](https://github.com/bcherny/json-schema-to-typescript/issues/428) from the main contributors to remove them, meanwhile, I think we can live with them.

### DIFF
--- a/typescript/thing-description/package.json
+++ b/typescript/thing-description/package.json
@@ -21,7 +21,7 @@
   },
   "homepage": "https://github.com/w3c/wot-scripting-api#readme",
   "scripts": {
-    "build": "json2ts schema/td-json-schema-validation.json > thing-description.d.ts"
+    "build": "json2ts --unreachableDefinitions --input schema/td-json-schema-validation.json  --output thing-description.d.ts"
   },
   "devDependencies": {
     "json-schema-to-typescript": "^10.1.4"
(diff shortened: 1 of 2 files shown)


danielpeintner:
Mhh, sad.... maybe we can do some command afterwards to remove those lines all together.

Anyhow, for the time being I am fine with it. Maybe it gets fixed soon.

danielpeintner (approved):

relu91:
Another try could be to write down a description of each rule, that should substitute the awful comment... but there are many rules 🤣

danielpeintner:
> Another try could be to write down a description of each rule, that should substitute the awful comment... but there are many rules 🤣

I don't think it is reasonable. Hence I would ahead
1. Use as is and hope it gets fixed soon
2. In the worst case a simple search&replace script *could* do the job since it is always the same text.

danielpeintner:
Scripting API 2022-01-31
* Can merge as is (maybe clean-up git history), Can be merged afterwards. 
* @danielpeintner creates issue to keep track (LINK TO BE ADDED)

relu91:
@danielpeintner ready to go

PR #381, opened 2022-02-10
feat: reintroduce "local" discovery method

JKRhb:
At the last meeting, we had a discussion regarding the reintroduction of the "local" discovery method to the specification, initiated by @mmccool. This PR provides a first starting point for further discussion, reusing text that was removed by https://github.com/w3c/wot-scripting-api/pull/316.

I think the direction of last meeting's discussion was that "local" should now also include multicast discovery and should return TDs from all supported sources that can be considered nearby in an automatic fashion. There is probably some distinction needed in this regard (for instance, where does the use of a CoRE Resource Directory fall into that is not part of the local network?) as well as the possibibility to configure which "submethods" provided by the implementation should actually be used and how.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 10, 2022, 8:19 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2FJKRhb%2Fwot-scripting-api%2F14cb907571a7a7c0a646b4873b59c3509b8749aa%2Findex.html%3FisPreview%3Dtrue)

```
📡 HTTP Error 429: https://rawcdn.githack.com/JKRhb/wot-scripting-api/14cb907571a7a7c0a646b4873b59c3509b8749aa/index.html
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/wot-scripting-api%23381.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3820,7 +3820,7 @@
             Request the underlying platform to start the discovery process, with the following parameters:
             <ul>
               <li>
-              If |filter|s <a href="#dom-thingfilter-method">|method|</a> is not defined or the value is `"local"`, use the widest discovery method supported by the underlying platform.
+              If |filter|s <a href="#dom-thingfilter-method">|method|</a> is not defined or the value is `"local"`, use all local discovery mechanisms supported by the underlying platform.
               </li>
               <li>
                 If |filter|s <a href="#dom-thingfilter-method">|method|</a> is `"directory"`, use the remote <a>Thing Directory</a> specified in |filter.url|.
--- a/index.html
+++ b/index.html
@@ -3751,6 +3751,9 @@
           "<dfn>directory</dfn>" for discovery based on a service provided by a
           <a>Thing Directory</a> specified by <a>ThingFilter</a>'s `url`.
         </li>
+        <li>
+          "<dfn>local</dfn>" for discovering <a>Thing</a>s defined in the same device or connected to the device by wired or wireless means.
+        </li>
       </ul>
     </section>
 
@@ -3761,7 +3764,7 @@
       </p>
       <pre class="idl">
         dictionary ThingFilter {
-          DiscoveryMethod method = "directory";
+          DiscoveryMethod method = "local";
           USVString? url;
           object? fragment;
           <!-- DOMString? query; -->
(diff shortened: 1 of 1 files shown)


zolkis:
At this point this needs more thinking IMHO.

I think we should we also `direct` and `directory`, together with the examples, and just _add_ `local`.

But at this point it should be specified better what implementations may or should do, how is it interpreted.
I would tie it to the security and privacy concept of locality.
There is "local to the runtime" concept, i.e. all features the wot scripting runtime can see and allowed to access/expose.
That may include (or not), access to "local network" (being wifi or ethernet), bluetooth (any or certain profiles), NFC, etc.
In the next turn, that may mean multiple protocols (with bindings), including multicast.

It should also be clear (multicast being involved) that this is an operational mode discovery, and not phase 1 discovery for provisioning.

The devices that are discovered should also be in operational mode, not in bootstrapping or provisioning mode.

For the latter use cases, we should use a different API entry point, tailored to bootstrapping/provisioning, that has very different security requirements and constraints.

All of these arguments don't change the API shape, only the algorithms and descriptive text.

zolkis:
This could be continued in an issue. The topic has been discussed in  #311, we need to reflect back, discuss the use cases for "local" discovery, etc.

This PR could be closed, or could remain WiP, but then it needs to be updated to reflect the latest spec changes, as it needs a separate method now - or a special filter? or a special directory? :)

The special (solution-specific) directory could always be an option, so the narrow waist principle could be applied, i.e. we don't necessarily need a separate standardized API method for this - unless we already know how to standardize such API. 

However, expressing developer intent ("I only need locally connected Things") might be a valid use case, so we should continue discussing.

relu91:
Scripting Call 30-01-2023:
- Close the PR
- Use it to open relevant issues

JKRhb:
Discussion could be continued in #396, for example.

PR #383, opened 2022-02-10
Pull before pushing in the CI

### LINKED ISSUE #382: Sync ThingModel action failed
see https://github.com/w3c/wot-scripting-api/actions/runs/1821091154

it seems the previous "sync-td" job seems to cause the issue? 

@relu91

relu91:
I would like to keep the two jobs parallel, therefore I've added a pull action before pushing to the upstream. In theory one of the two jobs might still push changes while the other has already pulled... but I want to give this a try. Fixes #382


danielpeintner:
Let's try it 👍

Issue #364, opened 2022-01-14
Discovery API - Alternatives


JKRhb:
When trying to implement the current discovery API I got the impression that its current design could be simplified by using a structure similar to the subscription/observation of events/properties. In particular, I had problems with implementing step two of the `next()` method:

> If the active property is true, wait until the discovery results internal slot is not empty.

Correct me if I am wrong, but it seemed to me as if this would only be possible by continuously checking the results, using a function like `setTimeout` or within a while loop (are there any implementation examples?). As this didn't seem that optimal to me, I tried out a callback based approach instead (similar to the `Subscription` interface and the `observeProperty`/`subscribeEvent` methods) which worked quite well so far.

I therefore wanted to suggest changing the signature of the `discover` method (or its potential successors  which are discussed in #222) to something like

```js
ThingDiscovery discover(DiscoveryListener listener, optional ThingFilter thingFilter = null);
```

where the `DiscoveryListener` function signature could look like the following:

```js
type DiscoveryListener = (thingDescription: ThingDescription) => void;
```

A minimal example could look like this:

```typescript
async function handleDiscovery(thingDescription: ThingDescription) {
    const consumedThing = await wot.consume(thingDescription);
    await consumedThing.invokeAction("foo");
}


wot.discover(handleDiscovery, { method: "direct", url: "coap://example.org/td" );
// Calling stop() can probably be omitted when the "direct" method is being used.
```

Here, the discovery process is initialized using the `direct` method. When a Thing Description has been retrieved successfully, the `handleDiscovery` function is called which consumes the TD and interacts with the Thing. If I am not mistaken, a similar approach (using a class called `Observable`) was already part of the specification some time ago but I still wanted to open this issue to discuss it as an alternative to the current `discover` method.

I am looking forward to your feedback!

danielpeintner:
Scripting Call 2022-01-17
* collect pros and cons
* collect/write examples?
* align with other API style, subscription mechanism
* in case of same power: difficulty of implementing one or the other (e.g, blocking until *new* data is available)

danielpeintner:
I have been thinking about this proposal and I like it. It well aligns with what we have for `Subscription`.

w.r.t. "power" I think one can achieve the same, e.g., 
* the discovery listener can `stop()`  if the desired result has been found

JKRhb:
@danielpeintner Thank you for the feedback! 

> w.r.t. "power" I think one can achieve the same, e.g.,
> 
[...]

PR #405, opened 2022-05-27
feat!: add new Discovery Web IDL definitions

JKRhb:
This is an attempt to create Web IDL definitions based on the example in https://github.com/w3c/wot-scripting-api/issues/364#issuecomment-1137703749. This is still a work in progress and far from mergeable, but we might already be able to use for discussing the details of the new API.

The most important changes are the addition of the [async iterable](https://webidl.spec.whatwg.org/#idl-async-iterable) keywords and the removal of the `next()` method (which is now available on the inherited `Symbol.asyncIterator` field). Assuming that the discovery process should start right away, I also removed the `start()` method. However, if it is good practice to have such a method in a web API, then it could of course still be included. 

Furthermore, as both the `direct` and the `directory` methods need URLs in order to work, I made the `filter` and its `url` field required. This could be either reverted if other discovery mechanisms should be supported, or sub-interfaces for the these two methods could be introduced, where `url` is in fact a required field.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-scripting-api/pull/405.html" title="Last updated on Oct 10, 2022, 11:41 AM UTC (099e1e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/405/e94e160...JKRhb:099e1e6.html" title="Last updated on Oct 10, 2022, 11:41 AM UTC (099e1e6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3762,8 +3762,8 @@
       </p>
       <pre class="idl">
         dictionary ThingFilter {
-          DiscoveryMethod method = "direct";
-          required USVString url;
+          DiscoveryMethod method = "directory";
+          USVString? url;
           object? fragment;
           <!-- DOMString? query; -->
         };
--- a/index.html
+++ b/index.html
@@ -3676,7 +3676,7 @@
     <pre class="idl">
       [SecureContext, Exposed=(Window,Worker)]
       interface ThingDiscovery {
-        constructor(ThingFilter filter);
+        constructor(optional ThingFilter filter = null);
         readonly attribute ThingFilter? filter;
         readonly attribute boolean active;
         readonly attribute boolean done;
--- a/index.html
+++ b/index.html
@@ -3691,9 +3691,6 @@
       The {{ThingDiscovery}} interface has a <code>next()</code> method and a <code>done</code> property, but it is not an <a href="https://tc39.es/proposal-async-iteration/">Iterable</a>. Look into <a href="https://github.com/w3c/wot-scripting-api/issues/177">Issue 177</a> for rationale.
     </p>
     <p>
-      The <dfn>discovery results</dfn> <a>internal slot</a> is an internal queue for temporarily storing the found {{ThingDescription}} objects until they are consumed by the application using the <a href="#the-next-method">next()</a> method. Implementations MAY optimize the size of this queue based on e.g. the available resources and the frequency of invoking the <a href="#the-next-method">next()</a> method.
-    </p>
-    <p>
       The <dfn>filter</dfn> property represents the discovery filter of type <a>ThingFilter</a> specified for the discovery.
     </p>
     <p>
@@ -3704,6 +3701,9 @@
     </p>
     <p>
       The <dfn>error</dfn> property represents the last error that occurred during the discovery process. Typically used for critical errors that stop discovery.
+    </p>
+    <p>
+      <!-- TODO: Mention AsyncIterator/AsyncIterable? -->
     </p>
 
     <section><h3>Constructing {{ThingDiscovery}}</h3>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3830,7 +3832,8 @@ <h3>The <dfn>start()</dfn> method</h3>
             Whenever a new <a>Thing Description</a> |td:ThingDescription| is discovered by the underlying platform, run the following sub-steps:
             <ol>
               <li>
-                <a href="https://fetch.spec.whatwg.org/#fetch-api">Fetch</a> |td| as a JSON object |json|. If that fails, set the <a href="#dom-thingdiscovery-error">error</a> property to {{SyntaxError}}, discard |td| and continue the discovery process.
+                Retrieve |td| as a JSON object using the <a>Protocol Binding</a> specified by the |filter|'s <dfn>url</dfn> property.
```
zolkis, line 3835: I would add a note that previously this was using Fetch, which remains a possibility, depending on binding.

index.html
```diff
@@ -3879,13 +3882,10 @@ <h3>The <dfn>next()</dfn> method</h3>
...
           <li>
-            If the <a href="#dom-thingdiscovery-active">active</a> property is `true`, wait until the <a>discovery results</a> <a>internal slot</a> is not empty.
-          </li>
-          <li>
-            If <a>discovery results</a> is empty and the <a href="#dom-thingdiscovery-active">active</a> property is `false`, set the <a href="#dom-thingdiscovery-done">done</a> property to `true` and reject |promise|.
+            If the <a href="#dom-thingdiscovery-active">active</a> property is `true`, call the `next` method on the Discovery object's Symbol.AsyncIterator.
           </li>
           <li>
-            Remove the first {{ThingDescription}} object |td| from <a>discovery results</a>.
+            If the `value` property in the yielded result from the Symbol.AsyncIterator is undefined and the `done` property is set to `true`, set the <a href="#dom-thingdiscovery-done">done</a> property to `true` and reject |promise|.
```
zolkis, line 3888: We need at least one example for how to handle async iterators in client code.

index.html
JKRhb: I added the line in 704191d2fa0925d3bbe632c57131cfd05583c934 for now, let me know if you see potential for improvement :)

index.html
JKRhb: I adjusted the first discovery example in 9646fb627c83b257cc671e70d4938ed0767d932d, I hope this points into the right direction.

JKRhb:
Made a couple of additional adjustments regarding the use of the `AsyncIterator`, removing mentions of the internal `results` slot. I think there is still some work to be done regarding the prose and the second example, but I suppose the current state can be used as a starting point for further discussion :) Therefore, I'd remove the draft status of the PR, indicating that it is ready for further reviews. (Thank you, @zolkis, for the feedback you already gave above!)

index.html
zolkis: Looks good!

index.html
zolkis: Fair.

index.html
```diff
@@ -3761,8 +3762,8 @@ <h3>The <dfn>ThingFilter</dfn> dictionary</h3>
       </p>
       <pre class="idl">
         dictionary ThingFilter {
-          DiscoveryMethod method = "directory";
-          USVString? url;
+          DiscoveryMethod method = "direct";
+          required USVString url;
```
zolkis, line 3766: About this, let's discuss in the call.

danielpeintner: Scripting call 2022-09-26
* Move out this change in another PR, keep this PR a syntactic change 
* `auto` method ?

JKRhb: Moved these lines to a new PR #434 now :)

zolkis (approved):

danielpeintner:
Scripting Call 2022-10-10
-> Changes Approved

Issue #408, opened 2022-06-13
Do we need an EventHandler? 


relu91:
In the [Handling Events](https://w3c.github.io/wot-scripting-api/#handling-events) flow, we are stating that if `data` is `null` or undefined the runtime should ask the internal slot `eventHandler` to return something for sending it to the network. 

However, I don't think we have such a use case: a caller does not know what to send for a particular event and defers the decision to the backend logic of an ExposedThing. 

This choice makes also it quite inconvenient to cover the use case described in #402. In other words, if an event has not data here's the code needed to implement it:
```js
var produced; // Produced Thing
produced.setEventHandler("noDataEvent", () => undefined);
// exposing
await produced.expose();
produced.emitEvent("noDataEvent");
```

Whereas if we just rely on the data passed on the `emitEvent` method we could simply:
```js
var exposed;
exposed.emitEvent("noDataEvent");
```

danielpeintner:
Call 2022-07-04
* recall use-case from TUM having this handler, @danielpeintner will look it up

egekorkan:
I am not 100% sure but this was about implementing a shadow thing or a proxy. I am proxying a local thing to the public network and before having this event handler, the proxy would subscribe to all events even if there is no consumer of the proxy interested in some of the events. With the event handler, when a subscription request comes to the proxy, it will subscribe in the event handler.

However, I think I am not seeing something since in my point of view, there is no difference in an event having data or not.

danielpeintner:
The issue @egekorkan is mentioning is https://github.com/w3c/wot-scripting-api/issues/211.

I "think" the issue is the same for events and observing property changes...

danielpeintner:
Call 2022-07-11
Maybe @egekorkan has a "picture/diagram" to show the problem... network configuration/setup. We were not 100% sure.
@egekorkan would you have time to join a call once?

egekorkan:
I can join next week. I think that I have confused it with eventsubscription handler

zolkis:
> agreed that there is no need for EventHandler

Let's record the justification in more detail, please. I wonder if we all mean the same things and use cases about this EventHandler.
- The context is ExposedThing, but specifying a customized (hooked) ConsumedThing side (how to handle events, for instance in case of Thing proxy implementations)
- In contrast, emitEvent() is the ExposedThing side
- Constructing an ExposedThing starts with a TD that defines event names
[...]

PR #423, opened 2022-08-08
Remove eventHandler

relu91:
As discussed #408 this PR removes references to `eventHandler` from the API.

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/423.html" title="Last updated on Sep 26, 2022, 11:40 AM UTC (fc07cc7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/423/f7eaed5...relu91:fc07cc7.html" title="Last updated on Sep 26, 2022, 11:40 AM UTC (fc07cc7)">Diff</a>


index.html
```diff
@@ -3398,6 +3322,12 @@ <h3>The <dfn>EventSubscriptionHandler</dfn> callback</h3>
                 Create an <a>Event</a> notification |response| according to the
                 <a>Protocol Bindings</a> from |data| and |subscriber|, including
                 its |options|.
+              </li>
+              <li>
+                  If |data| is `undefined`, assume that the notification |response| 
+                  will contain a `void` data payload.
```
zolkis, line 3328: I suggest reformulating to more exact prose, to something like

> If |data| is `undefined`, let |response|'s |data| be `undefined`

(needs to be done consistently over all occurrences of |response|

zolkis (approved):

index.html
danielpeintner: Call of 2022-08-08
@relu91 will work (provide) an alternative to make clear what `void` means..

danielpeintner (approved):
LGTM after resolving https://github.com/w3c/wot-scripting-api/pull/423#discussion_r939993623

index.html
```diff
@@ -3398,6 +3322,12 @@ <h3>The <dfn>EventSubscriptionHandler</dfn> callback</h3>
                 Create an <a>Event</a> notification |response| according to the
                 <a>Protocol Bindings</a> from |data| and |subscriber|, including
                 its |options|.
+              </li>
+              <li>
+                  If |data| is `undefined`, assume that the notification |response| 
+                  will contain a `empty` data payload as specified by <a>Protocol Bindings</a>.
```
zolkis, line 3328: This is too generic here. From an algorithm we need to point to exact steps or algorithms from another spec.
If the referred spec doesn't have that, we need to figure that out for the HTTP protocol binding and include the steps here. Let's track that in a separate issue.

danielpeintner, line 3328:
```suggestion
                  will contain an <emph>empty</emph> data payload as specified by <a>Protocol Bindings</a>. 
```

danielpeintner:
Scripting Call 2022-09-26
Agreed to merge

index.html
danielpeintner: see https://github.com/w3c/wot-scripting-api/issues/430

PR #434, opened 2022-10-10
feat: adjust `ThingFilter` and `ThingDiscovery` interfaces

JKRhb:
This PR contains the changes to the `ThingFilter` and `ThingDiscovery` interfaces that have originally been part of #405.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-scripting-api/pull/434.html" title="Last updated on Oct 10, 2022, 11:48 AM UTC (1319e56)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/434/e875c60...JKRhb:1319e56.html" title="Last updated on Oct 10, 2022, 11:48 AM UTC (1319e56)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3684,8 +3684,8 @@
     <pre class="idl">
       [SecureContext, Exposed=(Window,Worker)]
       interface ThingDiscovery {
-        constructor(optional ThingFilter filter = null);
-        readonly attribute ThingFilter? filter;
+        constructor(ThingFilter filter);
+        readonly attribute ThingFilter filter;
         readonly attribute boolean active;
         readonly attribute boolean done;
         readonly attribute Error? error;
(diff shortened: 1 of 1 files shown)


danielpeintner:
Does it make sense to state why you suggest both changes...

## w.r.t. `url`

I think I see the reasoning..

>  `direct` and the `directory` methods need URLs in order to work

## w.r.t. to require ThingFilter

Is it because one can pass in `null` anyway?
Did we check all algorithms?

JKRhb:
Thank you for your comment, @danielpeintner, you are right, the PR needs a bit more clarification 😅 The change of the ThingFilter was related to making the URL mandatory. I think the algorithms do indeed need to be adjusted, I did not manage to revisit them yet.

danielpeintner:
Closing since we move on with https://github.com/w3c/wot-scripting-api/issues/364

PR #424, opened 2022-08-08
Fix internal slots

relu91:
I noticed that we are not correctly referencing the internal slots we defined. Moreover, in `ExposedThing` we are using internal slots in the "old way". I'd like to fix this issue too.

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/424.html" title="Last updated on Dec 19, 2022, 6:28 PM UTC (b76dcee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/424/6fcfdb6...relu91:b76dcee.html" title="Last updated on Dec 19, 2022, 6:28 PM UTC (b76dcee)">Diff</a>

Issue #390, opened 2022-03-16
Passing Credentials to Discovery Methods


JKRhb:
One aspect that is currently missing when using the `discover` method is the ability to retrieve TDs that are behind some kind of authentication/authorization mechanism. For example, if a Thing (or a Directory) requires basic authentication to access its TD(s) you currently have no way to provide the username and password information. 

One possible solution I see for this would be to add an additional `security` parameter to the discover method which could contain both a `SecurityScheme` and an optional `Credentials` object.

With the current API design, this could look like this:

```ts
DiscoverySecurity = {
    SecurityScheme scheme,
    Credentials? credentials,
}


discover(optional ThingFilter filter = null, optional DiscoverySecurity security = null);
```

One challenge I see here is that you might need to define the format of credentials for each `SecurityScheme`, which might be especially challenging in the case of a `ComboSecurityScheme`. However, maybe the format of Credentials needs to be defined anyway in order to clarify their use with Scripting API implementations in general.

(It seems as if there have been related discussions in https://github.com/eclipse/thingweb.node-wot/issues/269 and https://github.com/w3c/wot-scripting-api/issues/216.)

zolkis:
Web APIs are discouraged to pass security tokens. It would be a more secure design to provision a runtime, or a script (context) with those credentials, for instance via a separate API (and permission levels) or other means (that can be protocol-specific). 

Let's do this based on use cases. We need strong justification to mix security provisioning with operational APIs.

JKRhb:
> It would be a more secure design to provision a runtime, or a script (context) with those credentials, for instance via a separate API (and permission levels) or other means (that can be protocol-specific).

Could the [Credentials Management API](https://developer.mozilla.org/en-US/docs/Web/API/Credential_Management_API) be used for this in the browser? Despite its experimental status, it seems as if there is already [good support](https://developer.mozilla.org/en-US/docs/Web/API/Credential_Management_API#browser_compatibility) (with IE being the only unsupported platform). In node-wot, the credentials store of the Servient could probably be used for this, I guess?

In any case, I think you either need a way to tell the discovery method which SecurityScheme should be used or it needs to be negotiated as part of the discovery process (e. g. if the Thing/Directory responds with an Unauthorized response code).

[...]

PR #441, opened 2022-11-28
Align the discovery API with the Discovery spec

zolkis:
Introduce separate methods for requesting to fetch a TD, for directory discovery and for generic discovery.
Rename ThingDiscovery to ThingDiscoveryProcess.
Remove |url| from ThingFilter and make it internal slot to ThingDiscoveryProcess. Remove |filter| from ThingDiscoveryProcess and make it internal slot to ThingDiscoveryProcess. Remove |active| from ThingDiscoveryProcess.
Define a generic discovery algorithm with meta-argument for generic or directory discovery.
Add the URL standard to the document header / xref.
Address #364

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-scripting-api/pull/441.html" title="Last updated on Dec 19, 2022, 12:32 PM UTC (27f9442)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/441/6fcfdb6...zolkis:27f9442.html" title="Last updated on Dec 19, 2022, 12:32 PM UTC (27f9442)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -212,7 +212,7 @@
           then instantiating a software stack that implements the <a>WoT Interface</a> specified by the <a>TD</a> in order to serve requests for accessing the exposed <a>Properties</a>, <a>Actions</a> and <a>Events</a>,
         </li>
         <li>
-           then eventually publishing the <a>Thing Description</a> (for instance to a <a>Thing Directory</a> directory for easier discovery).
+           then eventually publishing the <a>Thing Description</a> (for instance to a <a>TD Directory</a> directory for easier discovery).
         </li>
       </ul>
       This specification describes how to expose and consume <a>Thing</a>s by a script. Also, it defines a generic API for <a>Thing</a> discovery.
@@ -308,7 +308,7 @@
       <li>Discover <a>Thing</a>s running in the local <a>WoT Runtime</a>.</li>
       <li>Discover nearby <a>Thing</a>s, for instance connected by NFC or Bluetooth,
         or within a <a href="https://en.wikipedia.org/wiki/Geo-fence">geo-fence</a>.</li>
-      <li>Discover <a>Thing</a>s by sending a discovery request to a given <a>Thing Directory</a>.</li>
+      <li>Discover <a>Thing</a>s by sending a discovery request to a given <a>TD Directory</a>.</li>
       <li>Discover <a>Thing</a>s filtered by filters defined on <a>Thing Description</a>s</li>
       <li>Discover <a>Thing</a>s filtered by semantic queries.</li>
       <li>Stop or suppress an ongoing discovery process.</li>
@@ -751,7 +751,7 @@
       };
     </pre>
     <div>
-      Belongs to the <a>WoT Discovery</a> conformance class. Starts the discovery process that given a <a>Thing Directory</a> URL, will provide {{ThingDescription}} objects for <a>Thing Description</a>s that match an optional |filter:ThingFilter| argument of type {{ThingFilter}}. The method MUST run the following steps:
+      Belongs to the <a>WoT Discovery</a> conformance class. Starts the discovery process that given a <a>TD Directory</a> URL, will provide {{ThingDescription}} objects for <a>Thing Description</a>s that match an optional |filter:ThingFilter| argument of type {{ThingFilter}}. The method MUST run the following steps:
       <ol>
         <li>
           Return a {{Promise}} |promise:Promise| and execute the next steps [=in parallel=].
@@ -3736,7 +3736,7 @@
       <tr>
         <td>[[\url]]</td>
         <td>`undefined`</td>
-        <td>A {{URL}} representing the <a>Thing Directory</a> in a discovery.</td>
+        <td>A {{URL}} representing the <a>TD Directory</a> in a discovery.</td>
       </tr>
       </tbody>
     </table>
@@ -3798,12 +3798,12 @@
         <a>ThingFilter</a>, until it is standardized in the WoT Discovery
         task force. It represented a query string accepted by the implementation,
         for instance a SPARQL or JSON query. Support was to be implemented locally
-        in the <a>WoT Runtime</a> or remotely as a service in a <a>Thing Directory</a>.
+        in the <a>WoT Runtime</a> or remotely as a service in a <a>TD Directory</a>.
       </p>
       <p class="ednote">
         The |url:USVString| property was removed. It used to represent the target
         entity serving the discovery request, for instance the URL of a
-        <a>Thing Directory</a>, or the URL of a directly targeted <a>Thing</a>,
+        <a>TD Directory</a>, or the URL of a directly targeted <a>Thing</a>,
         but these are implemented by dedicated methods now.
       </p>
     </section>
@@ -3978,7 +3978,7 @@
         console.log("Found Thing Description for " + td.title);
       </pre>
       <p>
-        The next example finds {{ThingDescription}} objects of <a>Thing</a>s listed in a <a>Thing Directory</a> service. We set a timeout for safety.
+        The next example finds {{ThingDescription}} objects of <a>Thing</a>s listed in a <a>TD Directory</a> service. We set a timeout for safety.
       </p>
       <pre class="example" title="Discover Things via directory">
         let discovery = WOT.exploreDirectory("http://directory.wotservice.org");
--- a/index.html
+++ b/index.html
@@ -378,7 +378,7 @@
   
   <section> <h2>Terminology and conventions</h2>
     <p>
-      The generic WoT terminology is defined in [[!WOT-ARCHITECTURE]]: <dfn data-lt="Things">Thing</dfn>, <dfn data-lt="Thing Descriptions">Thing Description</dfn> (in short <dfn>TD</dfn>), <dfn>Partial TD</dfn>, <dfn>Web of Things</dfn> (in short <b><i>WoT</i></b>),  <dfn>WoT Interface</dfn>, <dfn>Protocol Bindings</dfn>, <dfn>WoT Runtime</dfn>, <dfn data-lt="consume|consume a TD|consuming a TD">Consuming a Thing Description</dfn>, <dfn>Thing Directory</dfn>, <dfn data-lt="Properties">Property</dfn>, <dfn data-lt="Actions">Action</dfn>, <dfn data-lt="Events|WoT-Event">Event</dfn>,
+      The generic WoT terminology is defined in [[!WOT-ARCHITECTURE]]: <dfn data-lt="Things">Thing</dfn>, <dfn data-lt="Thing Descriptions">Thing Description</dfn> (in short <dfn>TD</dfn>), <dfn>Partial TD</dfn>, <dfn>Web of Things</dfn> (in short <b><i>WoT</i></b>),  <dfn>WoT Interface</dfn>, <dfn>Protocol Bindings</dfn>, <dfn>WoT Runtime</dfn>, <dfn data-lt="consume|consume a TD|consuming a TD">Consuming a Thing Description</dfn>, <dfn>TD Directory</dfn>, <dfn data-lt="Properties">Property</dfn>, <dfn data-lt="Actions">Action</dfn>, <dfn data-lt="Events|WoT-Event">Event</dfn>,
       <a href="https://www.w3.org/TR/2020/WD-wot-thing-description11-20201124/#dataschema">
         <dfn>DataSchema</dfn></a>, <a href="https://www.w3.org/TR/2020/WD-wot-thing-description11-20201124/#form"><dfn data-lt="Forms">Form</dfn></a>,
         <a href="https://w3c.github.io/wot-thing-description/#securityscheme"><dfn>SecurityScheme</dfn></a>, <a href="https://w3c.github.io/wot-thing-description/#nosecurityscheme"><dfn>NoSecurityScheme</dfn></a> etc.
--- a/index.html
+++ b/index.html
@@ -3981,7 +3981,7 @@
         The next example finds {{ThingDescription}} objects of <a>Thing</a>s listed in a <a>TD Directory</a> service. We set a timeout for safety.
       </p>
       <pre class="example" title="Discover Things via directory">
-        let discovery = WOT.exploreDirectory("http://directory.wotservice.org");
+        let discovery = await WOT.exploreDirectory("http://directory.wotservice.org");
         setTimeout( () => {
             discovery.stop();
             console.log("Discovery stopped after timeout.");
@@ -4001,7 +4001,7 @@
         to the WOT runtime, including local Things, if any is available.
       </p>
       <pre class="example" title="Discover Things in a network">
-        let discovery = WOT.discover();
+        let discovery = await WOT.discover();
         setTimeout( () => {
             discovery.stop();
             console.log("Stopped open-ended discovery");
--- a/index.html
+++ b/index.html
@@ -52,7 +52,7 @@
             ]
           },
         ],
-        xref: ["web-platform", "streams", "wot-architecture", "wot-thing-description", "wot-binding-templates", "html", "infra", "ecmascript"],
+        xref: ["web-platform", "streams", "wot-architecture", "wot-thing-description", "wot-binding-templates", "html", "infra", "url", "ecmascript"],
         localBiblio: {
           "WOT-ARCHITECTURE" : {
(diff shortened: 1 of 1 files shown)


zolkis:
@relu91 @danielpeintner @JKRhb PTAL
you can also contribute by pushing commits to this PR.

relu91:
from 05/12/2022:
- still, need to be reviewed.
- @danielpeintner implementing `requestThingDescription` is hard cause it should support a vast range of URL schemes.

index.html
```diff
@@ -702,27 +706,120 @@ <h3>Validating an ExposedThingInit</h3>
   <section> <h3>The <dfn>discover()</dfn> method</h3>
     <pre class="idl">
       partial namespace WOT {
-        ThingDiscovery discover(optional ThingFilter filter = null);
+        Promise&lt;ThingDiscoveryProcess&gt; discover(optional ThingFilter filter = null);
```
JKRhb, line 709: Sorry if I have missed it, but I suppose here we would use the `ExtendedThingFilter`, right?

index.html
```diff
@@ -3697,149 +3786,181 @@ <h3>The <dfn>ThingFilter</dfn> dictionary</h3>
...
-      <p>
-        The <dfn>url</dfn> property represents the URL of the target entity serving the discovery request, for instance the URL of a <a>Thing Directory</a> (if <a href="#dom-thingfilter-method">method</a> is {{DiscoveryMethod/[["directory"]]}}),
-        or the URL of a directly targeted <a>Thing</a> (if
-        <a href="#dom-thingfilter-method">method</a> is {{DiscoveryMethod/[["direct"]]}})
+      <p class = "ednote">
+        The |query:DOMString| property was temporarily removed from
+        <a>ThingFilter</a>, until it is standardized in the WoT Discovery
+        task force. It represented a query string accepted by the implementation,
+        for instance a SPARQL or JSON query. Support was to be implemented locally
+        in the <a>WoT Runtime</a> or remotely as a service in a <a>Thing Directory</a>.
```
JKRhb, line 3801: IIRC, Thing Directories have been renamed to Thing Description Directories.

```suggestion
        in the <a>WoT Runtime</a> or remotely as a service in a <a>Thing Description Directory</a>.
```

index.html
```diff
@@ -3697,149 +3786,181 @@ <h3>The <dfn>ThingFilter</dfn> dictionary</h3>
...
       </p>
-      <p>
-        The <dfn>fragment</dfn> property represents a template object used for matching property by property against discovered <a>Thing</a>s.
+      <p class="ednote">
+        The |url:USVString| property was removed. It used to represent the target
+        entity serving the discovery request, for instance the URL of a
+        <a>Thing Directory</a>, or the URL of a directly targeted <a>Thing</a>,
```
JKRhb, line 3806:
```suggestion
        <a>Thing Description Directory</a>, or the URL of a directly targeted <a>Thing</a>,
```

index.html
```diff
@@ -3852,51 +3973,47 @@ <h3>Discovery Examples</h3>
...
-        let discovery = new ThingDiscovery({ method: "direct" });
-        for await (let td of discovery) {
-          console.log("Found Thing Description for " + td.title);
-        }
+        let url = "https://mythings.com/thing1";
+        let td = await WOT.requestThingDescription(url);
+        console.log("Found Thing Description for " + td.title);
       </pre>
       <p>
         The next example finds {{ThingDescription}} objects of <a>Thing</a>s listed in a <a>Thing Directory</a> service. We set a timeout for safety.
```
JKRhb, line 3981:
```suggestion
        The next example finds {{ThingDescription}} objects of <a>Thing</a>s listed in a <a>Thing Description Directory</a> service. We set a timeout for safety.
```

index.html
```diff
@@ -702,27 +706,120 @@ <h3>Validating an ExposedThingInit</h3>
...
+
+  <section> <h3>The <dfn>exploreDirectory()</dfn> method</h3>
+    <pre class="idl">
+      partial namespace WOT {
+        Promise&lt;ThingDiscoveryProcess&gt; exploreDirectory(USVString url,
+            optional ThingFilter filter = null);
+      };
+    </pre>
+    <div>
+      Belongs to the <a>WoT Discovery</a> conformance class. Starts the discovery process that given a <a>Thing Directory</a> URL, will provide {{ThingDescription}} objects for <a>Thing Description</a>s that match an optional |filter:ThingFilter| argument of type {{ThingFilter}}. The method MUST run the following steps:
```
JKRhb, line 754:
```suggestion
      Belongs to the <a>WoT Discovery</a> conformance class. Starts the discovery process that given a <a>Thing Description Directory</a> URL, will provide {{ThingDescription}} objects for <a>Thing Description</a>s that match an optional |filter:ThingFilter| argument of type {{ThingFilter}}. The method MUST run the following steps:
```

index.html
```diff
@@ -3697,149 +3786,181 @@ <h3>The <dfn>ThingFilter</dfn> dictionary</h3>
...
+                of |discovery|.
               </li>
-            </ul>
-          </li-->
-          <!-- <li>
-            Create the <a>discovery results</a> <a>internal slot</a> for storing discovered {{ThingDescription}} objects.
-          </li> -->
+              <li>
+                Request the underlying platform to start the directory discovery
+                passing |url| and |filter| to it.
```
JKRhb, line 3837: Is this enough, or do we need to be more specific here? Given that a directory is described by a Thing Description (derived from [this Thing Model](https://w3c.github.io/wot-discovery/#directory-api-spec)), I assume we could also specify that this TD needs to be consumed and the `things` property needs to be read?

index.html
```diff
@@ -3697,149 +3786,181 @@ <h3>The <dfn>ThingFilter</dfn> dictionary</h3>
...
+        but these are implemented by dedicated methods now.
       </p>
-      <!--p>
-        The <dfn>query</dfn> property represents a query string accepted by the implementation, for instance a SPARQL or JSON query. Support may be implemented locally in the <a>WoT Runtime</a> or remotely as a service in a <a>Thing Directory</a>.
-      </p-->
     </section>
 
     <section>
-      <h3>The <dfn>start()</dfn> method</h3>
+      <h3>The <dfn data-lt="generic-discovery">generic discovery</dfn> algorithm</h3>
```
JKRhb, line 3812: A question partly related to https://github.com/w3c/wot-scripting-api/issues/390: Can/should we specify the runtime's behavior if it encounters an "Unauthorized" response? This relates to the [Security Bootstrapping](https://w3c.github.io/wot-discovery/#exploration-secboot) feature in the Discovery specification.

index.html
```diff
@@ -3852,51 +3973,47 @@ <h3>Discovery Examples</h3>
...
       </pre>
-      <!--p>
-        The next example is for an open-ended multicast discovery, which likely won't complete soon (depending on the underlying protocol), so stopping it with a timeout is a good idea. It will likely deliver results one by one.
+      <p>
+        The next example is for a generic discovery, by any means provisioned
+        to the WOT runtime, including local Things, if any is available.
       </p>
       <pre class="example" title="Discover Things in a network">
-        let discovery = new ThingDiscovery({ method: "multicast" });
+        let discovery = WOT.discover();
```
JKRhb, line 4004: With the new API shape, I suppose we need an `await` here, right?

```suggestion
        const discovery = await WOT.discover();
```

index.html
```diff
@@ -3852,51 +3973,47 @@ <h3>Discovery Examples</h3>
...
       <p>
         The next example finds {{ThingDescription}} objects of <a>Thing</a>s listed in a <a>Thing Directory</a> service. We set a timeout for safety.
       </p>
       <pre class="example" title="Discover Things via directory">
-        let discoveryFilter = {
-          method: "directory", // default value, no need to specify
-          url: "http://directory.wotservice.org"
-        };
-        let discovery = new ThingDiscovery(discoveryFilter);
+        let discovery = WOT.exploreDirectory("http://directory.wotservice.org");
```
JKRhb, line 3984:
```suggestion
        const discovery = await WOT.exploreDirectory("http://directory.wotservice.org");
```

JKRhb (commented):
Thanks again for your work, @zolkis! See a couple of comments below :)

index.html
zolkis: I checked and formally actually made no sense (yet) to extend ThingFilter. We usually simply use prose in the algorithms to describe which properties are used from `filter`. It doesn't matter if someone provides an object that has extra properties, as long as there are the ones that are needed.

index.html
zolkis: We discussed in the call that we might need an algorithm here, imported/adapted from the Discovery spec. 
However, if there are differences in protocol bindings, it is not clear where these algorithms need to be specified: in the Discovery spec, or in Bindings, or in Scripting. 
It feels awkward to specify in detail what to do with the content of a TD... that has not been in the scope of Scripting API.
This is one thing to figure out in this PR - for now, leaving it open for comments.

index.html
zolkis: First, if such a step is explicit, should be in the algorithms of the methods, not in the meta-algorithms.
But this is still a variant of " if not allowed".
Dialogs spawned by the runtime (or other local decisions/mechanisms/policies) are not in the scope of the API spec. That is something that happens in the underlying platform, but might have an effect here under the common clause of "not allowed". For privacy reasons, specs usually don't detail how exactly this failed.

JKRhb: I see, then we can consider this question resolved :)

relu91:
I've not properly reviewed all the changes and new algorithms, but I probably spotted a simple issue. Internal slots should be referred to using the correct syntax as I was trying to fix in #424.

zolkis:
Quite many changes in last commit, needs complete re-review.
@relu91 @danielpeintner @JKRhb

index.html
```diff
@@ -170,7 +176,7 @@
     </p>
     <p>
       Scripting is an optional building block in WoT and it is typically used in gateways or browsers that are able to run a <a>WoT Runtime</a> and
-      <a href="https://github.com/w3c/wot-scripting-api/tree/master/applications/script-manager">script management</a>, providing a convenient way to extend WoT support to new types of endpoints and implement WoT applications such as <a href="https://github.com/w3c/wot-scripting-api/tree/master/applications/thing-directory">Thing Directory</a>.
+      <a href="https://github.com/w3c/wot-scripting-api/tree/master/applications/script-manager">script management</a>, providing a convenient way to extend WoT support to new types of endpoints and implement WoT applications such as <a href="https://github.com/w3c/wot-scripting-api/tree/master/applications/thing-directory">TD Directory</a>.
```
danielpeintner, line 179: More a question: Why this change.. discovery seems to speak still about "Thing Directory".
https://w3c.github.io/wot-discovery/#exploration-td-type-thingdirectory

index.html
```diff
@@ -212,7 +218,7 @@
           then instantiating a software stack that implements the <a>WoT Interface</a> specified by the <a>TD</a> in order to serve requests for accessing the exposed <a>Properties</a>, <a>Actions</a> and <a>Events</a>,
         </li>
         <li>
-           then eventually publishing the <a>Thing Description</a> (for instance to a <a>Thing Directory</a> directory for easier discovery).
+           then eventually publishing the <a>Thing Description</a> (for instance to a <a>TD Directory</a> directory for easier discovery).
```
danielpeintner, line 221: same as above

index.html
```diff
@@ -308,7 +314,7 @@
       <li>Discover <a>Thing</a>s running in the local <a>WoT Runtime</a>.</li>
       <li>Discover nearby <a>Thing</a>s, for instance connected by NFC or Bluetooth,
         or within a <a href="https://en.wikipedia.org/wiki/Geo-fence">geo-fence</a>.</li>
-      <li>Discover <a>Thing</a>s by sending a discovery request to a given <a>Thing Directory</a>.</li>
+      <li>Discover <a>Thing</a>s by sending a discovery request to a given <a>TD Directory</a>.</li>
```
danielpeintner, line 317: Same again

index.html
```diff
@@ -3600,94 +3733,86 @@ <h2>ExposedThing Examples</h2>
...
+      interface ThingDiscoveryProcess {
         constructor(optional ThingFilter filter = null);
-        readonly attribute ThingFilter? filter;
-        readonly attribute boolean active;
         readonly attribute boolean done;
         readonly attribute Error? error;
-        undefined start();
-        Promise&lt;ThingDescription&gt; next();
         undefined stop();
         async iterable&lt;ThingDescription&gt;;
```
danielpeintner, line 3752: Nit: double semicolon

danielpeintner (approved):
I think it looks good. I added question about naming...

index.html
zolkis: @relu91 please check... IIRC we discussed in the call that we use TD Directory instead of Thing Directory.
I am fine with either, personally I like Thing Directory more.

index.html
zolkis: Old error, not in this PR, but will fix with other things if I need something else fixing, too.

danielpeintner: Ahh, I see 🙈

index.html
JKRhb: In the terminology, "Thing Directory" has been renamed to TDD: https://w3c.github.io/wot-discovery/#exploration-directory I suppose that this change is partly motivated by the abbreviations (TD vs. TDD), as otherwise the directory could be confused with the Thing Description.

JKRhb: You are right though, @danielpeintner, that the renaming in the Discovery spec does not seem to be that consistent...

zolkis:
I suggest we leave the "TD Directory" for now.

index.html
```diff
@@ -702,27 +712,150 @@ <h3>Validating an ExposedThingInit</h3>
...
-          Return |discovery|.
+          Set |discovery|'s {{ThingDiscoveryProcess/[[filter]]}} to |filter:ThingFilter|.
+        </li>
+        <li>
+          Set |discovery|'s {{ThingDiscoveryProcess/[[url]]}} to `undefined`.
+        </li>
+        <li>
+          If |filter| is not supported by the implementation, or if discovery
+          cannot be started by the underlying platform,
+          reject |promise| with {{NotSupportedError}} and abort these steps.
```
relu91, line 743: I would split the two cases into two different errors. Throwing `NotSupportedError` for a condition in which the discovery cannot be started feels unspecific. Shouldn't be better to throw https://webidl.spec.whatwg.org/#operationerror ?

index.html
```diff
@@ -702,27 +712,150 @@ <h3>Validating an ExposedThingInit</h3>
...
+        </li>
+        <li>
+          If invoking this method is not allowed for the current scripting
+          context for security reasons, reject |promise| with a {{SecurityError}}
+          and abort these steps.
+        </li>
+        <li>
+          If fetching a <a>Thing Description</a> is not supported by the
+          implementation, reject |promise| with {{NotSupportedError}} and
+          abort these steps.
```
relu91, line 843: Here is also a good place to put the same error condition as specified in the `exploreDirectory` method:
> if the underlying
              implementation cannot support the <a>Protocol Binding</a>
              indicated by |url|

relu91 (approved):
There are two minor comments below, other than that I think we have a good draft that we can use to start experimenting.

index.html
zolkis: Right. And there could be even 3 errors, 
- an error if the filter is not supported at all,  (`NotSupportedError`?)
- another if the filter contained an error, (`DataError`?)
- and yet another if discovery could not be started (`OperationError`?)
It's also a balance between being specific and giving fingerprinting differentiation. 
If we need one error, maybe `OperationError` would be fine.

index.html
zolkis: Mmm, in this case I think a `NotSupportedError` is straightforward, but we can discuss it.

index.html
zolkis: Maybe `NotSupportedError` and `OperationError` would be enough. I think the information that a filter is not supported, or filters in general are not supported, is not directly usable for fingerprinting, though correlations with HW capability, impl version etc could be statistically made. It's a policy decision that can even alter request by request or filter by filter.

The question is what to do if there is an error in a filter. Should the algoritm
- ignore the filter and provide everything it gets
- fail. 
Currently it fails and I think it should.

JKRhb (approved):

index.html
relu91: From 19/12/2022:
- @zolkis if we go into detail there would be a lot of other clauses, it is better to stay generic.

Issue #402, opened 2022-05-20
Emitting no data for events


relu91:
Currently, emit event signature force to send a kind of InteractionInput which might be a DataSchema. Therefore, in my understanding, if I don't have any data for the event and since InteracitonInput is *not* optional, I should send `null` and event payload. Two questions:
- In JS it might be more idiomatic to have InteractionInput as optional -> the runtime will understand that it should not send any value
- "null" is still a valid JSON value so in principle, there might be corner cases where the developer would actually want to send "null" as a network value. How do we handle such use cases?

JKRhb:
Leaving out the optional parameter would produce an `undefined` value, so maybe the developer could either do that or explicitly pass `undefined`? Then the two cases (`null` vs. `undefined`) could be distinguished.

relu91:
The problem is that `undefined` is not a valid  `InteractionInput` value. Moreover, in WebIDL optional means most of the time that the value can be  `undefined`. So my proposal is to update the  `eventEmit` signature to:
```webidl
Promise<undefined> emitEvent(DOMString name, optional InteractionInput data )
```
which in TS translates to
```ts
async emitEvent(name: string, data?: InteractionInput)
```

zolkis:
Sounds good, make a PR.
Technically [InteractionInput](https://w3c.github.io/wot-scripting-api/#the-interactioninput-type) is either a `ReadableStream` or any... which can be `undefined`.
However, it would be good to explicitly express optionality in the API.

relu91:
> Technically [InteractionInput](https://w3c.github.io/wot-scripting-api/#the-interactioninput-type) is either a ReadableStream or any... which can be undefined.

Good catch I was using TS and in our type definitions DataSchemaValue (which in WebIDL is any) is more restrictive. See https://github.com/w3c/wot-scripting-api/blob/main/typescript/scripting-api/index.d.ts#L87 . I'll make a PR but now I wondering what we should do about that DataSchemaValue definition... I think we can discuss this in another Issue.

danielpeintner:
Call 2022-05-30
* add `optional` AND check steps algorithm

danielpeintner:
Call 2022-07-04
* we miss PR that changes
https://github.com/w3c/wot-scripting-api/blob/9294bc8899a2c578f3c7f12b32a65dfd1b9be500/typescript/scripting-api/index.d.ts#L291
to
` emitEvent(name: string, data?: InteractionInput): void;`
* check algorithm

PR #442, opened 2022-12-12
fix: emitEvent() method allows data to be optional

danielpeintner:
resolves https://github.com/w3c/wot-scripting-api/issues/402

***
<a href="https://pr-preview.s3.amazonaws.com/danielpeintner/wot-scripting-api/pull/442.html" title="Last updated on Dec 14, 2022, 12:37 PM UTC (c323d8a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/442/6fcfdb6...danielpeintner:c323d8a.html" title="Last updated on Dec 14, 2022, 12:37 PM UTC (c323d8a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3369,7 +3369,7 @@
           </li>
           <li>
             Make a request to the underlying platform to emit an <a>Event</a>
-            with optionally |data|. Call the <a href="#handling-events">handling events</a>
+            with optional |data|. Call the <a href="#handling-events">handling events</a>
             steps.
           </li>
         </ol>
--- a/index.html
+++ b/index.html
@@ -2398,7 +2398,7 @@
         ExposedThing setEventUnsubscribeHandler(DOMString name,
                 EventSubscriptionHandler handler);
         Promise&lt;undefined&gt; emitEvent(DOMString name,
-                InteractionInput data);
+                optional InteractionInput data = null);
 
         Promise&lt;undefined&gt; expose();
         Promise&lt;undefined&gt; destroy();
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -3369,7 +3369,7 @@ <h3>The <dfn>EventSubscriptionHandler</dfn> callback</h3>
           </li>
           <li>
             Make a request to the underlying platform to emit an <a>Event</a>
-            with |data|. Call the <a href="#handling-events">handling events</a>
+            with optionally |data|. Call the <a href="#handling-events">handling events</a>
```
zolkis, line 3372: optional

danielpeintner: Thanks, fixed.

relu91:
I checked the changes, but I have a question. In [handling events](https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/442/6fcfdb6...danielpeintner:1156f8c.html#handling-events) we are saying "if `data` is `undefined` then ... " however in the WebIDL we are setting the default to `null`, in my understanding this means that if `emitEvent` function is called with just one parameter (i.e. `emitEvent("test")`) data is `null`. Should we change the default to `undefined`? Note that some payload formats allow for `null` as a value (see [JSON](https://www.rfc-editor.org/rfc/rfc4627.html#section-2.1)), this means that in practice a runtime could still interpret  `emitEvent("test")` as the intention to send `null` on the wire, making it unintuitive.

zolkis:
No, we must not initialize to `undefined`, that's the default anyway. Null initialization makes sense if the algorithm requires it, but in that case it doesn't.

zolkis:
If we don't need null initialization, or init with default values, we just leave it without init.

danielpeintner:
> If we don't need null initialization, or init with default values, we just leave it without init.

I don't think we need initialization for `emitEvent(...)`. On the contrary, it seem to cause issues as @relu91 alluded.

Moreover, I would argue we should also remove all the other `null` initialization like
`optional InteractionOptions options = null`
to
`optional InteractionOptions options`
etc.

Opinions?

danielpeintner:
I removed the initialization for emitEvent(...) and created a new issue https://github.com/w3c/wot-scripting-api/issues/445 to discuss the other remaining methods with null initialization.

relu91 (approved):
LGTM, we can go on :)

zolkis (approved):

PR #450, opened 2023-02-06
Fix #417: add optional data argument to emitPropertyChange()

### LINKED ISSUE #417: emitPropertyChange does not take low-level event apis into account
With the changes at #407 , the `emitPropertyChange` calls the read method on the property. However, it might be that I am subscribed to a temperature value over a serial connection and I would emit the temperature value as a property change once I detect it. Reading the property value would be redundant and might even give the wrong value if the temperature changes very quickly.

The gpio library at npm to control GPIOs of a Raspberry Pi has such a subscription API at https://www.npmjs.com/package/gpio#eventemitter . In this case, I would like to write the following code:
```js
gpio4.on("change", function(val) {
   // value will report either 1 or 0 (number) when the value changes
   console.log(val)
   myThing.emitPropertyChange(val);
});
```

With the current changes, the code would be:
```js
gpio4.on("change", function(val) {
   // value will report either 1 or 0 (number) when the value changes
   console.log(val)
   myThing.emitPropertyChange( );
   //internally, the following will happen via the readhandler
   let propertyVal = gpio4.get() //this is the same as val above if the gpio has not changed value
   return propertyVal
});
```

I find this to be not exactly optimal. The freedom to supply a value would be good and this can be the default behavior.

zolkis:
***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/wot-scripting-api/pull/450.html" title="Last updated on Feb 7, 2023, 12:08 PM UTC (77ba098)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/450/2708c45...zolkis:77ba098.html" title="Last updated on Feb 7, 2023, 12:08 PM UTC (77ba098)">Diff</a>


index.html
```diff
@@ -3012,25 +3018,35 @@ <h3>The <dfn>PropertyReadHandler</dfn> callback</h3>
...
+                If |handler| is `null`, abort these steps.
+              </li>
+              <li>
+                Let |promise| be the result of invoking |handler| with `null`.
+              </li>
+              <li>
+                 If |promise| rejects, abort these steps.
+              </li>
+              <li>
+                Otherwise let |data| be the value |promise| resolves with.
```
danielpeintner, line 3043: Not sure, but for me it would read better like this 

Otherwise let |data| be the value **the** |promise| resolves with.

danielpeintner (approved):
The changes look good to me.

I added a minor comment about how to phrase one step but it I am not 100% sure...

index.html
zolkis: |promise| is a named variable declared before, so it doesn't need "the"

zolkis: Maybe we can reformulate the whole sentence, suggestions welcome.

zolkis:
Note that I updated the algorithm with up-to-date spec prose. We need to do the same for other algorithms as well.
Some steps are still not exact, we can handle that later at the big sweep.

index.html
danielpeintner: It's fine by me... it just reads somehow _strange_

zolkis: It's been changed.

index.html
```diff
@@ -2988,62 +2989,86 @@ <h3>The <dfn>PropertyReadHandler</dfn> callback</h3>
 
     <section> <h3>The <dfn>emitPropertyChange()</dfn> method</h3>
       <div>
-        Takes the |name:string| argument, denoting a <a>Property</a> name.
+        Takes the |name:string| argument, denoting a <a>Property</a> name, and
+        optionally the |data:InteractionInput| argument.
         Triggers emitting a notification to all observers of the specified
-        <a>Property</a> with the data obtained from the observe handler or the
+        <a>Property</a> with the data provided by the |data| argument, or if not
+        provided, it is obtained by the observe handler or the by
```
relu91, line 2996: *Note*: the observe handler is not mentioned later on in the algorithm. Also, I think the observe handler now has another function: it is called when somebody observes the property.

relu91 (changes requested):

index.html
zolkis: This is not related to this PR. Let's take a look at that problem in a different issue and PR.

relu91: Ok, let's address this in a follow-up PR.

relu91 (approved):

Issue #454, opened 2023-02-20
Use BibTex references also in the text


relu91:
Using the diff of this https://github.com/w3c/wot-scripting-api/pull/451, you can see that there are hardcoded references in the specification text. We should use BibTex refs as much as possible.

PR #456, opened 2023-02-20
Use bib references in text instead of URLs

danielpeintner:
Note1: I used [data-cite](https://respec.org/docs/#data-cite) to be able to use fragment identifiers. I know [xref](https://respec.org/docs/#xref) is the *new* recommended way of linking to terms in other specs. However, xref does not seem to support fragment identifiers. 

Note2: I think using bib references everywhere makes it very easy and simple to update to a new version of a spec. Hence it reduces costs and eliminates issues like that a URL has been forgotten to be updated.

Note3: There are more URLs we could update, but I am not sure which xref key to use for the following URLs:
* https://fetch.spec.whatwg.org/ for #fetch-api and other
* https://streams.spec.whatwg.org for #readablestream-get-a-reader and others
* https://html.spec.whatwg.org/multipage/origin.html for #same-origin-domain and others

fixes https://github.com/w3c/wot-scripting-api/issues/454

### DIFF
--- a/index.html
+++ b/index.html
@@ -486,7 +486,7 @@
       Represents a <a>Thing Description</a> (<a>TD</a>) as
       <a data-cite="WOT-TD#">defined</a> in
       [[!WOT-TD]]. It is expected to be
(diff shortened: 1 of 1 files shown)


zolkis (approved):

relu91:
Call 27/02/2023: PR approved.

TODOs: About Note3 please use https://respec.org/xref.

Issue #445, opened 2022-12-14
null initialization necessary for optional method arguments?


danielpeintner:
> > If we don't need null initialization, or init with default values, we just leave it without init.
> 
> I don't think we need initialization for `emitEvent(...)`. On the contrary, it seem to cause issues as @relu91 alluded.
> 

Yes exactly, I'd remove the `null` initialization here. About the other places, let's talk about it in an issue (?) or maybe directly on a dedicated PR.

_Originally posted by @relu91 in https://github.com/w3c/wot-scripting-api/issues/442#issuecomment-1350877086_

danielpeintner:
As mention in https://github.com/w3c/wot-scripting-api/pull/442#issuecomment-1350504813 we have many methods that uses null initialization
* Promise<InteractionOutput> readProperty(DOMString propertyName,
                              optional InteractionOptions options = null);

* Promise<InteractionOutput> invokeAction(DOMString actionName,
                              optional InteractionInput params = null,
                              optional InteractionOptions options = null);

* ThingDiscovery discover(optional ThingFilter filter = null);
* ...

I don't think they are needed.

The TypeScript destinations don't have that either, see for example
https://github.com/w3c/wot-scripting-api/blob/6fcfdb65fbec2af5cc5b6243ee09d7ae69388a25/typescript/scripting-api/index.d.ts#L114

I can provide a PR if you are okay with removing them...

danielpeintner:
Note: I created https://github.com/w3c/wot-scripting-api/pull/461 but this PR does not remove null initialization since WebIDL seems to require it (see _more_ information in PR itself).

zolkis:
For the record: Web IDL requires that optional dictionaries are initialized by the `default dictionary` that is the `{}` constant. See the relevant Note in https://webidl.spec.whatwg.org/#idl-constants

PR #461, opened 2023-03-01
add `undefined` where missing in prose

danielpeintner:
I added in some places `undefined` where I thought it is missing (besides `null`)

Note: This PR **does not** remove `= null` for optional arguments since WebIDL seems to require them.

image

fixes https://github.com/w3c/wot-scripting-api/issues/445

### DIFF
--- a/index.html
+++ b/index.html
@@ -975,7 +975,7 @@
         </li>
         <li>
           If the value of the |data| property is not a {{ReadableStream}} or if
-          |dataUsed| is `true`, or if |form| is `null` or if |schema| or its
+          |dataUsed| is `true`, or if |form| is `null` or `undefined` or if |schema| or its
           |type| are `null` or `undefined`,
           reject |promise| with {{NotReadableError}} and abort these steps.
         </li>
@@ -3170,7 +3170,7 @@
             Otherwise, if there is a default write handler provided by the implementation, let |handler| be that.
           </li>
           <li>
-            Otherwise, if |handler| is `null`, send back a {{NotSupportedError}}
+            Otherwise, if |handler| is `null` or `undefined`, send back a {{NotSupportedError}}
             with the reply and abort these steps.
           </li>
           <li>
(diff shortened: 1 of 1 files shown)


relu91 (approved):
I'm okay with the changes. About WebIDL this means that an implementation like node-wot should put always a default value in the optional arguments that are `null`. what is the advantage of having set to `null `instead of `undefined`?

danielpeintner:
> About WebIDL this means that an implementation like node-wot should put always a default value in the optional arguments that are `null`.

I don't think it has any effect on implementations like node-wot since it should thread optional arguments already properly.
I believe WebIDL just wants so see _any_ default value. @zolkis ?

> what is the advantage of having set to `null `instead of `undefined`?

Good point. I was asking myself the same.
Changing all appearances of `= null` to `= undefined` seemed reasonable to me as well and would be aligned with what the TypeScript definitions do. Anyhow, somehow WebIDL does not like that :-(

image

zolkis (approved):

danielpeintner:
Call 2023-03-06
* add empty objects for WebIDL **and** TS
* @danielpeintner

danielpeintner:
Note:

Changing the following TypeScript code

https://github.com/w3c/wot-scripting-api/blob/14ff4ac1ce7d12ae50c1c893c8686ec304070499/typescript/scripting-api/index.d.ts#L13

to 

`export function discover(filter?: ThingFilter = {}): ThingDiscovery;`

causes the following error:
"A parameter initializer is only allowed in a function or constructor implementation.ts(2371)"

I am not 100% sure if it is possible to match TypeScript with WebIDL in all cases.

I am fine with changing `null` to `{}` etc in WebIDL but I guess we should leave TS as is.
I am even fine with merging this PR as is and leave `= null` in WebIDL.

zolkis (changes requested):
In WebIDL let's go with the practice required there: initialize optional dictionaries with `{}`.
The Web IDL spec notes the default dictionary {} can be used as a constant - but that likely doesn't help TypeScript unless there is a special, matching constant for that.

danielpeintner:
> In WebIDL let's go with the practice required there: initialize optional dictionaries with `{}`.

implemented by https://github.com/w3c/wot-scripting-api/pull/461/commits/d8063ebe2023e728b2aaa97b8e1ac59b3ab2a727

zolkis (approved):

relu91 (approved):
It would be cool to align TS but as described above it seems we have some limitations. Let's maybe keep track of this in an issue and merge this PR.

Issue #475, opened 2023-05-15
Pubrules checker hangs for current published version at 35/37


relu91:
Pubrules checker hangs for me at 35/37

https://www.w3.org/pubrules/?url=https%3A%2F%2Fhtmlpreview.github.io%2F%3Fhttps%3A%2F%2Fgithub.com%2Fw3c%2Fwot-scripting-api%2Fblob%2F46082af917909826a8c5dd49387f89e484078e04%2Freleases%2Fnote2%2Findex.html&profile=NOTE&validation=simple-validation&informativeOnly=false&echidnaReady=false&patentPolicy=pp2020

_Originally posted by @danielpeintner in https://github.com/w3c/wot-scripting-api/issues/473#issuecomment-1525186239_

PR #477, opened 2023-05-15
2023 note2 updates2

danielpeintner:
[Preview](https://htmlpreview.github.io/?https://raw.githubusercontent.com/w3c/wot-scripting-api/adc0aa95db7feb27c27fa253974b3eb44e8025f5/releases/note2/index.html)

NOTE: All (some?) of the fixes should go into the main note as well in a follow-up PR

fixes https://github.com/w3c/wot-scripting-api/issues/475

### DIFF
--- a/releases/note2/index.html
+++ b/releases/note2/index.html
@@ -1514,7 +1514,7 @@ interface <a class="internalDFN idlID" data-link-type="interface" href="#dom-int
       implementation.
     </p>
     <figure id="scripting-read">
-      <img src="images/scripting-read-data.png" title="Reading data" width="100%">
+      <img src="images/scripting-read-data.png" title="Reading data" style="width:100%;">
       <figcaption><a class="self-link" href="#scripting-read">Figure <bdi class="figno">1</bdi></a> <span class="fig-title">Data structures used when reading data</span></figcaption>
     </figure>
     <p>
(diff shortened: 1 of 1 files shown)


zolkis (approved):

danielpeintner:
@ashimura the remaining issue with duplicate id "idl-def-wot" is coming from ReSpec. We in our html do not create this ID, please have a look at https://github.com/w3c/wot-scripting-api/blob/main/index.html. You will **not** find "idl-def-wot"..

It comes from the WebIdl where we use `partial namespace WOT ` like in

```
partial namespace WOT {
  Promise<ConsumedThing> consume(ThingDescription td);
};
```

or

```
partial namespace WOT {
  Promise<ExposedThing> produce(ExposedThingInit init);
};
```

or

```
partial namespace WOT {
  Promise<ThingDiscoveryProcess> exploreDirectory(USVString url,
      optional ThingFilter filter = {});
};
```


How can we move on?
@zolkis can we define it differently?

zolkis:
Those definitions are correct, according to https://webidl.spec.whatwg.org/#dfn-partial-namespace

danielpeintner:
I created https://github.com/w3c/respec/issues/4447, maybe there is a simple solution/fix for it.

PR #502, opened 2023-09-29
refactor: move WoT definitions up in a single place

danielpeintner:
Note: gets rid of partial namespace definitions since they don't seem to work

fixes validation issues mentioned in https://github.com/w3c/wot-scripting-api/pull/477 and https://github.com/w3c/respec/issues/4447

***
<a href="https://pr-preview.s3.amazonaws.com/danielpeintner/wot-scripting-api/pull/502.html" title="Last updated on Sep 29, 2023, 6:47 AM UTC (83db8c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/502/78a28a6...danielpeintner:83db8c2.html" title="Last updated on Sep 29, 2023, 6:47 AM UTC (83db8c2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -552,15 +552,22 @@
   <pre class="idl">
     [SecureContext, Exposed=(Window,Worker)]
     namespace WOT {
-      // methods defined in UA conformance classes
+      // WoT Consumer conformance class
+      Promise&lt;ConsumedThing&gt; consume(ThingDescription td);
+	  
+      // WoT Producer conformance class
+      Promise&lt;ExposedThing&gt; produce(ExposedThingInit init);
+	  
+      // WoT Discovery conformance class
+      Promise&lt;ThingDiscoveryProcess&gt; discover(optional ThingFilter filter = {});
+      Promise&lt;ThingDiscoveryProcess&gt; exploreDirectory(USVString url,
+        optional ThingFilter filter = {});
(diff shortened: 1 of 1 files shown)


zolkis:
Doesn't change any Web IDL, only the prose.

Having conformance classes has been quite a consensus in the group. Not everyone must implement discovery, for instance.
With this change, the normative of the spec (currently a Note) applies to all functionality becomes normative together at once.

I don't see the benefit of this and should be discussed / agreed in an issue before a PR.

danielpeintner:
> I don't see the benefit of this and should be discussed / agreed in an issue before a PR.

We got an email from @ashimura that unfortunately our current spec is causing issues for the publication, see https://github.com/w3c/respec/issues/4447

I just a try to find a solution. Anyhow, @ashimura is trying something else and report back.

zolkis:
Oh, so we have the `partial interface` causing a problem? Might be a ReSpec issue?
There's no problems with partial declarations with bikeshed, for instance.

relu91:
Call 02/10:
- @relu91 we should fix the respec
- @zolkis let's keep discussion conformance classes in #506, the publication has been fixed by @ashimura.

danielpeintner:
As decided in https://github.com/w3c/wot-scripting-api/issues/506 we keep using conformance classes -> closing PR

PR #513, opened 2023-10-23
feat!(typescript): align TypeScript definitions with Discovery API

JKRhb:
In the context of https://github.com/eclipse-thingweb/node-wot/pull/864, I noticed that the TypeScript definitions are not taking into account the latest changes to the Scripting APIs discovery API (see https://github.com/w3c/wot-scripting-api/pull/405 and https://github.com/w3c/wot-scripting-api/pull/441). This PR should adjust the types accordingly – let me know if I missed anything :)

### DIFF
--- a/typescript/scripting-api/index.d.ts
+++ b/typescript/scripting-api/index.d.ts
@@ -61,6 +61,7 @@ declare namespace WoT {
      * controlling the discovery process. 
      */
     export interface ThingDiscoveryProcess extends AsyncIterable<ThingDescription> {
+        new(filter?: ThingFilter);
         filter?: ThingFilter;
         done: boolean;
         error?: Error;
--- a/typescript/scripting-api/index.d.ts
+++ b/typescript/scripting-api/index.d.ts
@@ -12,7 +12,7 @@ declare namespace WoT {
      *
      * @param filter represents the constraints for discovering Things as key-value pairs
      */
-    export function discover(filter?: ThingFilter): ThingDiscoveryProcess;
+    export function discover(filter?: ThingFilter): Promise<ThingDiscoveryProcess>;
 
     /**
      * Starts the discovery process that, given a TD Directory {@link url}, will
--- a/typescript/scripting-api/index.d.ts
+++ b/typescript/scripting-api/index.d.ts
@@ -51,10 +51,6 @@ declare namespace WoT {
      */
     export interface ThingFilter {
         /**
-         * The query field represents a query string accepted by the implementation, for instance a SPARQL query.
-         */
-        query?: string;
-        /**
          * The fragment field represents a template object used for matching against discovered Things.
          */
         fragment?: object;
--- a/typescript/scripting-api/index.d.ts
+++ b/typescript/scripting-api/index.d.ts
@@ -64,8 +64,6 @@ declare namespace WoT {
         filter?: ThingFilter;
         done: boolean;
         error?: Error;
-        start(): void;
-        next(): Promise<ThingDescription>;
         stop(): void;
     }
 
--- a/typescript/scripting-api/index.d.ts
+++ b/typescript/scripting-api/index.d.ts
@@ -6,11 +6,30 @@ type DeepPartial<T> = T extends object ? {
 declare namespace WoT {
 
     /**
-     * Starts the discovery process that will provide ConsumedThing 
-     * 
+     * Starts the discovery process that will provide {@link ThingDescription}
+     * objects for Thing Descriptions that match an optional {@link filter}
+     * argument of type {@link ThingFilter}.
+     *
      * @param filter represents the constraints for discovering Things as key-value pairs
      */
-    export function discover(filter?: ThingFilter): ThingDiscovery;
+    export function discover(filter?: ThingFilter): ThingDiscoveryProcess;
+
+    /**
+     * Starts the discovery process that, given a TD Directory {@link url}, will
+     *  provide {@link ThingDescription} objects for Thing Descriptions that
+     * match an optional {@link filter} argument of type {@link ThingFilter}.
+     *
+     * @param url URL pointing to a TD Directory.
+     * @param filter represents the constraints for discovering Things as key-value pairs
+     */
+    export function exploreDirectory(url: string, filter?: ThingFilter): Promise<ThingDiscoveryProcess>;
+
+    /**
+     * Requests a {@link ThingDescription} from the given {@link url}.
+     *
+     * @param url The URL to request the Thing Description from.
+     */
+    export function requestThingDescription(url: string): Promise<ThingDescription>;
 
     /**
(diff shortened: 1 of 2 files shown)


typescript/scripting-api/index.d.ts
```diff
@@ -32,15 +51,7 @@ declare namespace WoT {
...
-        /**
-         * The url field represents additional information for the discovery method, such as the URL of the target entity serving the discovery request, such as a Thing Directory or a Thing.
-         */
-        url?: string;
-        /**
-         * The query field represents a query string accepted by the implementation, for instance a SPARQL query. 
+         * The query field represents a query string accepted by the implementation, for instance a SPARQL query.
          */
         query?: string;
         /**
```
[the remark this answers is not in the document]
JKRhb: Thank you :) I wasn't entirely sure about this one because in the source HTML file `query` is just commented out, but I suppose removing makes more sense to perform a proper alignment.

typescript/scripting-api/index.d.ts
```diff
@@ -49,25 +60,12 @@ declare namespace WoT {
...
      * The ThingDiscovery object is constructed given a filter and provides the properties and methods
      * controlling the discovery process. 
      */
-    export interface ThingDiscovery {
+    export interface ThingDiscoveryProcess extends AsyncIterable<ThingDescription> {
         filter?: ThingFilter;
-        active: boolean;
         done: boolean;
         error?: Error;
         start(): void;
```
danielpeintner, line 67: There is no _good_ match ..
* stop vs start etc

image

typescript/scripting-api/index.d.ts
```diff
@@ -6,11 +6,30 @@ type DeepPartial<T> = T extends object ? {
...
     /**
-     * Starts the discovery process that will provide ConsumedThing 
-     * 
+     * Starts the discovery process that will provide {@link ThingDescription}
+     * objects for Thing Descriptions that match an optional {@link filter}
+     * argument of type {@link ThingFilter}.
+     *
+     * @param filter represents the constraints for discovering Things as key-value pairs
+     */
+    export function discover(filter?: ThingFilter): ThingDiscoveryProcess;
```
danielpeintner, line 15: I think it needs to be the returning a Promise 

```suggestion
    export function discover(filter?: ThingFilter): Promise<ThingDiscoveryProcess>;
```

typescript/scripting-api/index.d.ts
JKRhb: Thanks! I think I missed `start` and `next` when adjusted these.

JKRhb: Should be resolved by 3be7d2a740dc0b7b343215f40648dfe7baf0659c

typescript/scripting-api/index.d.ts
```diff
@@ -31,47 +50,20 @@ declare namespace WoT {
...
-        "multicast"
-    }
-
     /**
      * The ThingDiscovery object is constructed given a filter and provides the properties and methods
      * controlling the discovery process. 
      */
-    export interface ThingDiscovery {
+    export interface ThingDiscoveryProcess extends AsyncIterable<ThingDescription> {
         filter?: ThingFilter;
```
danielpeintner, line 64: I have one more question about `filter` since it should be part of the constructor.

image


For `ConsumedThing` we have a **_constructor_** as well and we do not have it as a variable there...
What do you think should we do here?

typescript/scripting-api/index.d.ts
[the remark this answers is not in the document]
danielpeintner: Maybe it comes back but for the time being I think we should remove it.

typescript/scripting-api/index.d.ts
JKRhb: Hmmm, good question, I was wondering the same, but I think in TypeScript interfaces, it is not possible to define a constructor, is it?

JKRhb: My impression was that TypeScript and WebIDL interfaces simlpy differ in this regard, but I am really not sure here.

danielpeintner: Mhh, I just think we need to align both cases...

Having said that, since we cannot model constructors in TypeScript I think we should not mention/list `filter` at all here. 
Anyhow, I would like to hear comments from others as well. @relu91 @zolkis others?

zolkis: @danielpeintner I don't get the problem, as there are constructors in TypeScript. 
But the spec should be clear about mapping WebIDL object, ECMAScript class and TypeScript [class](https://www.typescriptlang.org/docs/handbook/2/classes.html).

danielpeintner: Classes have constructors, of course. What @JKRhb and I have been looking for are constructors on interfaces. Apparently even those exist but I have not seen them before...

https://www.typescriptlang.org/docs/handbook/interfaces.html#difference-between-the-static-and-instance-sides-of-classes

JKRhb: > Apparently even those exist but I have not seen them before...

Me neither, actually, but using `new` seems to work :) I wonder if we should adjust the other TypeScript interfaces like this as well then?

danielpeintner: I think so.. we are using "constructor" for 2 interfaces. `ThingDiscoveryProcess` and `ConsumedThing`

JKRhb: In order to make things a bit easier to track, I opened a separate PR for the `ConsumedThing` interface: https://github.com/w3c/wot-scripting-api/pull/515

danielpeintner:
Call Nov 06
* @zolkis  need to check whether we need constructor. Factory to use for async --> remove constructor for both cases, ConsumedThing as well
* @zolkis should link "old" PR when we changed the spec itself (in first comment)
* @JKRhb removing constructors is fine for me
* @danielpeintner update version number of API, https://github.com/w3c/wot-scripting-api/blob/dc3a512bca7e858e6adc58be2b6762dd04194da7/typescript/scripting-api/package.json#L3

JKRhb:
The PR should now be updated as discussed during today's meeting :) Let me know if there is anything that should be changed additionally.

danielpeintner (approved):
LGTM

danielpeintner:
@JKRhb May I ask you to update the version to "0.8.0-SNAPSHOT.29" since "0.8.0-SNAPSHOT.28" has been used by https://github.com/w3c/wot-scripting-api/pull/522

JKRhb:
> @JKRhb May I ask you to update the version to "0.8.0-SNAPSHOT.29" since "0.8.0-SNAPSHOT.28" has been used by #522

Done :)

danielpeintner:
Scripting Call Nov 20
* updates look good
* group decided to merge as is

PR #515, opened 2023-10-24
feat(typescript): add constructor to ConsumedThing interface

JKRhb:
Corresponding with the discussion in https://github.com/w3c/wot-scripting-api/pull/513#discussion_r1369907606, this PR adds a constructor parameter to the TypeScript definition of the `ConsumedThing` interface.

In this context, I noticed that we might want to also add a constructor to both the WebIDL and TypeScript definitions of the `ExposedThing` interface.


danielpeintner:
Closed in favor of #517

Issue #512, opened 2023-10-23
Add perl script (for fixing duplicate IDs) to master/release folder


danielpeintner:
Based on the outcome of https://github.com/w3c/wot-scripting-api/issues/506 we should look into whether we want to store the perl script / readme for fixing the duplicate id issues.

see https://github.com/w3c/wot-scripting-api/pull/503/ and https://github.com/w3c/wot-scripting-api/pull/504

danielpeintner:
Call Nov 6
* @ashimura should update README documenting how to use this script
* @danielpeintner looks into creating a PR, for https://github.com/w3c/wot-scripting-api/blob/main/releases (copy content from TD repo)

Issue #480, opened 2023-05-22
Follow up issue for respec error using WEB IDL partial namespace


relu91:
I created https://github.com/w3c/respec/issues/4447, maybe there is a simple solution/fix for it.

_Originally posted by @danielpeintner in https://github.com/w3c/wot-scripting-api/issues/477#issuecomment-1554487654_

Also look at this comment:  https://github.com/w3c/wot-scripting-api/pull/477#issuecomment-1554299220

danielpeintner:
Keep open until https://github.com/w3c/respec/issues/4447 has been fixed. Once fixed we can partly remove things from https://github.com/w3c/wot-scripting-api/pull/518 (remove script and step 5)

danielpeintner:
There seems to be a ReSpec fix on the way for solving duplicates id errors when generating HTML for WebIDL using partial namespace, see https://github.com/speced/respec/issues/4447 (or https://github.com/speced/respec/commit/a015f323f7e6e61762b3c60fc4817bb69c92dd11)

EDIT: I tried with https://validator.w3.org/ and the errors still exist (might take some time till ReSpec gets updated)

PR #518, opened 2023-11-06
refactor: add readme and perl script for publication/release

danielpeintner:
fixes https://github.com/w3c/wot-scripting-api/issues/512

README.md based on https://github.com/w3c/wot-thing-description/blob/main/publication/README.md


relu91 (approved):

danielpeintner:
Scripting API Call Nov 20
* Group is ok
* Once the _root_ issue is fixes (see https://github.com/w3c/wot-scripting-api/issues/480) we can remove the perl script and the step 5

PR #523, opened 2023-11-20
chore: increase root package-lock.json

danielpeintner:
see https://github.com/w3c/wot-scripting-api/pull/513/files


danielpeintner:
I will merge right away since it is just about TypeScript and no real change..

PR #524, opened 2023-11-30
Update text and examples regarding the fetching of TDs

JKRhb:
In the current Editor's draft, there are a couple of leftovers from previous versions where the `requestThingDescription` had not been introduced yet. This PR updates the text, incorporates the latest changes, and also removes an outdated appendix that explains why there is no fetch method anymore (which has now de facto been reintroduced). Maybe some text regarding the validation of TDs could be added/moved elsewhere, though.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-scripting-api/pull/524.html" title="Last updated on Dec 4, 2023, 3:19 PM UTC (329213a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/524/6482a8c...JKRhb:329213a.html" title="Last updated on Dec 4, 2023, 3:19 PM UTC (329213a)">Diff</a>

PR #525, opened 2023-11-30
Use simple single and double quotes in code examples

JKRhb:
Resolves a general minor issue raised by @danielpeintner in https://github.com/w3c/wot-scripting-api/pull/524#discussion_r1410389930. Maybe we could also consider using one type of quote everywhere, but that could probably be handled automatically by something like prettier.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-scripting-api/pull/525.html" title="Last updated on Nov 30, 2023, 11:06 AM UTC (08f31d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/525/521e7d1...JKRhb:08f31d7.html" title="Last updated on Nov 30, 2023, 11:06 AM UTC (08f31d7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2400,7 +2400,7 @@
         };
 
         try {
-          // subscribe to property change for “temperature”
+          // subscribe to property change for "temperature"
           await thing.observeProperty("temperature", async (data) => {
             try {
               console.log("Temperature changed to: " + await data.value());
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4262,8 +4262,8 @@ <h2>API design rationale</h2>
           let res = await fetch(‘https://td.my.com/lock-00123’);
```
JKRhb, line 4262: Note that I did not change this line since it is already covered by #524.

relu91 (approved):

PR #534, opened 2024-01-22
fix(InteractionOutput): don't require schema.type in value function

JKRhb:
At the moment, the `value` function of the `InteractionOutput` interface requires the `type` field of its `schema` to be set. In a discussion over at https://github.com/eclipse-thingweb/node-wot/issues/1221, we identified this as a problem for several different scenarios, including the `exloreDirectory` method, as the `things` property of the Thing Description Directory API specification does not contain a `type` field itself, as there are two different types that can be obtained from the property (namely, an array or an object of TDs).

Therefore, this PR proposes changing the algorithm to not throw anymore when the field should be `null` or `undefined`, making it usable in more scenarios.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-scripting-api/pull/534.html" title="Last updated on Feb 27, 2024, 3:08 PM UTC (b38cf77)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/534/d0c5f8c...JKRhb:b38cf77.html" title="Last updated on Feb 27, 2024, 3:08 PM UTC (b38cf77)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1093,7 +1093,7 @@
               steps on |payload| and |subSchema|.
             </li>
             <li>
-              If none or more than two of these runs do not throw, throw
+              If none or more than one of these runs do not throw, throw
               {{TypeError}} and stop.
             </li>
           </ol>
--- a/index.html
+++ b/index.html
@@ -1093,7 +1093,8 @@
               steps on |payload| and |subSchema|.
             </li>
             <li>
-              If all of these runs throw, throw {{TypeError}} and stop.
+              If more than two of these runs do not throw, throw {{TypeError}}
+              and stop.
             </li>
           </ol>
         </li>
--- a/index.html
+++ b/index.html
@@ -1061,13 +1061,6 @@
           If |schema| is `null` or `undefined`, return `undefined`.
         </li>
         <li>
-          <!--
-            TODO: Should we just continue with the rest of the algorithm here?
-          -->
-          If |schema|.|default| is not |undefined| and |payload| is undefined,
-          let |payload| be the value of |schema|.|default|.
-        </li>
-        <li>
           If |schema|.|const| is not |undefined| and |schema|.|const| does not equal |payload|, throw {{TypeError}} and stop.
         </li>
         <li>
--- a/index.html
+++ b/index.html
@@ -976,8 +976,8 @@
         </li>
         <li>
           If |this|.|data| is not a {{ReadableStream}} or if
-          |dataUsed| is `true`, or if |form| is not an {{object}} or if |schema| or its
-          |type| are `null` or `undefined`, then
+          |dataUsed| is `true`, or if |form| is not an {{object}} or if |schema|
+          is `null` or `undefined`, then
           [=reject=] |promise| with {{NotReadableError}} and stop.
         </li>
         <li>
--- a/index.html
+++ b/index.html
@@ -1058,6 +1058,46 @@
       To run the <a>check data schema</a> steps on |payload| and |schema:object|,
       <ol>
         <li>
+          If |schema| is `null` or `undefined`, return `undefined`.
+        </li>
+        <li>
+          <!--
+            TODO: Should we just continue with the rest of the algorithm here?
+          -->
+          If |schema|.|default| is not |undefined| and |payload| is undefined,
+          let |payload| be the value of |schema|.|default|.
+        </li>
+        <li>
+          Let |const| be |schema|.|const|.
+          If |const| is not |undefined|,
+          <ol>
+            <li>
+              <!-- TODO: Should we already return here? -->
+              If |const| equals |payload|, return |payload|.
+            </li>
+            <li>
+              If |const| does not equal |payload|, throw {{TypeError}} and stop.
+            </li>
+          </ol>
+        </li>
+        <li>
+          If |schema|.|enum| is not |undefined| and none of the elements in
+          |schema|.|enum| equal |payload|, throw {{TypeError}} and stop.
+        </li>
+        <li>
+          Let |oneOf| be |schema|.|oneOf|.
+          If |oneOf| is not |undefined|,
+          <ol>
+            <li>
+              For each |subSchema| in |oneOf|, run the <a>check data schema</a>
+              steps on |payload| and |subSchema|.
+            </li>
+            <li>
+              If all of these runs throw, throw {{TypeError}} and stop.
+            </li>
+          </ol>
+        </li>
+        <li>
           Let |type| be |schema|.|type|.
         </li>
         <li>
--- a/index.html
+++ b/index.html
@@ -1068,17 +1068,7 @@
           let |payload| be the value of |schema|.|default|.
         </li>
         <li>
-          Let |const| be |schema|.|const|.
-          If |const| is not |undefined|,
-          <ol>
-            <li>
-              <!-- TODO: Should we already return here? -->
-              If |const| equals |payload|, return |payload|.
-            </li>
-            <li>
-              If |const| does not equal |payload|, throw {{TypeError}} and stop.
-            </li>
-          </ol>
+          If |schema|.|const| is not |undefined| and |schema|.|const| does not equal |payload|, throw {{TypeError}} and stop.
         </li>
         <li>
           If |schema|.|enum| is not |undefined| and none of the elements in
--- a/index.html
+++ b/index.html
@@ -1093,8 +1093,8 @@
               steps on |payload| and |subSchema|.
             </li>
             <li>
-              If more than two of these runs do not throw, throw {{TypeError}}
-              and stop.
+              If none or more than two of these runs do not throw, throw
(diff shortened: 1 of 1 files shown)


danielpeintner (approved):
Given the discussion we had in https://github.com/eclipse-thingweb/node-wot/issues/1221, and the arguments provided it makes sense to _soften_ the restriction a bit

relu91 (changes requested):
Sorry, I probably failed to express this correctly, but we **need** the data schema in the value function, and without it the value function does not make any sense... it is like mapping from A to B without B. Hence, as I said in [this comment](https://github.com/eclipse-thingweb/node-wot/issues/1221#issuecomment-1900057552), the algorithm should say:

> ... or if |schema| is `null` or is not a valid schema then reject...

Problem is that we need to define what is a valid schema for us. We got around this because so far for scripting API the schema did not have complex root types (like oneOf, allOf etc.). See also this section of how to [use the schema](https://www.w3.org/TR/wot-scripting-api/#the-check-data-schema-algorithm), it is very simplified but in reality (also in node-wot) schemas are much more complex than that.

JKRhb:
Very good points, @relu91, thank you. You are right, this definitely requires more work, especially when it comes to the "check data schema" algorithm. Could we maybe refer to something already defined by JSON Schema here?

zolkis (approved):

index.html
```diff
@@ -976,8 +976,8 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
         </li>
         <li>
           If |this|.|data| is not a {{ReadableStream}} or if
-          |dataUsed| is `true`, or if |form| is not an {{object}} or if |schema| or its
-          |type| are `null` or `undefined`, then
+          |dataUsed| is `true`, or if |form| is not an {{object}} or if |schema|
+          is `null` or `undefined`, then
```
JKRhb, line 980: For reference, this is (more or less) the change proposed by @relu91 in https://github.com/w3c/wot-scripting-api/pull/534#pullrequestreview-1838591689. The question here, also raised by him, is how we should ensure that the schema is valid – do we need to define a separate algorithm for that?

```suggestion
          |dataUsed| is `true`, or if |form| is not an {{object}} or if |schema|
          is `null` or `undefined` or not a valid schema, then
```

JKRhb: Oh, yeah, I now remember that we also discussed this briefly on Friday during the Thingweb meeting... Then I suppose we should keep this step as is, right?

relu91:
Call 05/02:
- @JKRhb relates to the todo item of DataMapping, so that's why it is hard to fix without knowing the result of the discussion
- @danielpeintner we can try a best-effort approach and suggest proper schema validation in an Editor's note
- @relu91 agrees, we can modify the algorithm to support the following keywords that are defined in the TD:
   1. `oneOf`
   2. `const`
   3. `enum`
   4. `default`  ? (related to https://github.com/w3c/wot-scripting-api/issues/537)
- @ashimura It would be important to describe the use cases of each one of these keywords. @relu91 I agree let's open an issue before merging this PR.

JKRhb:
I've now tried to incorporate `oneOf`, `enum`, `default`, and `const` into the _check data schema_ algorithm in 0f00764cce6f813ba8c168bab682eef1d6791037. There are at least two aspects of the algorithm that we could discuss in the call later today – I marked these with `TODO` comments in the HTML document.

JKRhb:
Regarding the use cases, please find a brief first attempt (that definitely needs more work) for the four terms below:

# oneOf

Relevant when a device described by a TD offers two different representations of the same resource (e.g., a property or an action output). A concrete example is the `things` property in the Thing Description Directory API, where – depending on the URI variable `format` – the directory can return TDs either as an array or a "collection", i.e. as part of a JSON object.

# const

As a developer, I might want to specify that a Thing always returns a certain value when an action is performed. A simple example could be a value like "Success". Combined with `oneOf`, we could describe that an interaction (in this case an action) always either returns "Success" or "Operation failed" like so:

```js
{
  "actions": {
    "exampleAction": {
      "output": {
        "oneOf": [
          {
            "const": "Success"
          },
          {
            "const": "Operation failed"
          }
        ]
      }
    }
  }
}
```

# default

As a developer, I want to save bandwidth, especially in resource-constrained environments. Therefore, I want to be able to specify a default value that has to be assumed by a TD consumer if the response sent after an operation has been completed does not contain a value.

# enum

As a developer, I want to constrain the values that are valid as an interaction output to a set of values. This could be used to describe the allowed states of a smart light ("on" or "off"), for example.

index.html
zolkis: "not a valid schema" needs an algorithm

index.html
```diff
@@ -1057,6 +1057,47 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
+          Let |const| be |schema|.|const|.
+          If |const| is not |undefined|,
+          <ol>
+            <li>
+              <!-- TODO: Should we already return here? -->
+              If |const| equals |payload|, return |payload|.
+            </li>
+            <li>
+              If |const| does not equal |payload|, throw {{TypeError}} and stop.
+            </li>
```
JKRhb, line 1080: I think the first condition could also be removed, as this way the other checks within the algorithm will also be applied to the payload (which would also uncover schemes that are unsatisfiable).

index.html
```diff
@@ -1057,6 +1057,47 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
+          If |oneOf| is not |undefined|,
+          <ol>
+            <li>
+              For each |subSchema| in |oneOf|, run the <a>check data schema</a>
+              steps on |payload| and |subSchema|.
+            </li>
+            <li>
+              If none or more than two of these runs do not throw, throw
+              {{TypeError}} and stop.
+            </li>
```
JKRhb, line 1098: I now adjusted this step of the algorithm as can be seen here. As this term checks whether there is exactly one matching `subSchema`, 0 or > 2 successful runs would indicate an invalid payload here.

JKRhb: (Just realized that this needs to be 0 or > 1 run instead, updated in 2a3365f0e1e1d1586d289287a1862d2b72bc4dd8.

index.html
```diff
@@ -1057,6 +1057,47 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
+          If |const| is not |undefined|,
+          <ol>
+            <li>
+              <!-- TODO: Should we already return here? -->
+              If |const| equals |payload|, return |payload|.
+            </li>
+            <li>
+              If |const| does not equal |payload|, throw {{TypeError}} and stop.
+            </li>
+          </ol>
```
JKRhb, line 1081: Corresponding with https://github.com/w3c/wot-scripting-api/pull/534/files#r1502130414, this could become something like this:

```suggestion
          If |schema|.|const| is not |undefined| and |schema|.|const| does not equal |payload|, throw {{TypeError}} and stop.
```

danielpeintner: FYI: JSON Schema is very unpredictable in a way. The following schema is valid 

```json
{
 "type": "number",
 "const": "abc"
}
```

but there is no possible valid input. Anyhow, the schema itself is valid according to some testing with https://jsonschemalint.com.

Anyhow, your proposal makes sense to me.

JKRhb: Thank you for your feedback, @danielpeintner! I will then apply the suggestion above :)

index.html
```diff
@@ -1057,6 +1057,37 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
       <ol>
+        <li>
+          If |schema| is `null` or `undefined`, return `undefined`.
+        </li>
+        <li>
+          <!--
+            TODO: Should we just continue with the rest of the algorithm here?
+          -->
+          If |schema|.|default| is not |undefined| and |payload| is undefined,
+          let |payload| be the value of |schema|.|default|.
```
JKRhb, line 1068: I now had another look into the JSON Schema Draft 7 specification (which is the basis for the `DataSchema` vocabulary in the TD) and noticed that `default` is actually only an "Annotation" (see [here](https://json-schema.org/draft-07/draft-handrews-json-schema-validation-01#annotations) and [here](https://json-schema.org/draft-07/draft-handrews-json-schema-validation-01#rfc.section.10)) that is supposed to be a hint used in documentation but not for validation.

On the one hand, I think that should be clarified in the TD document (as it is only implied by the reference to JSON Schema, which is only informative, though). On the other hand, I think that means we can simply remove this step from the algorithm as the default value is not supposed to be provided by the consumer if no payload is given.

JKRhb: I now removed the step for `default` in b38cf776c5bd41be7bdc81f2ed4ea6701934160a.

relu91:
From 04/03:
- @relu91 changes look good (+1 for removing default), they might not be perfect but they can refined later
- @danielpeintner approves the changes

Issue #535, opened 2024-01-22
Dedicated methods for DNS-SD and CoRE Link-Format discovery?


JKRhb:
Experimenting with both [DNS-SD](https://www.w3.org/TR/2023/REC-wot-discovery-20231205/#introduction-dns-sd-sec) and discovery via the [CoRE Link-Format](https://www.w3.org/TR/2023/REC-wot-discovery-20231205/#introduction-core-rd-sec) in dart_wot, I got the impression that it could make sense to add dedicated discovery methods for both approaches and maybe even replace the generic `discover` method in this context.

As the two approaches belong to the introduction phase, the new methods could be used solely for discovering URLs pointing to Things, which could be wrapped in objects containing a `url` and a `type` (such as `Thing` or `Directory`). The results could then be followed up with either the `requestThingDescription` or the `exploreDirectory` method. As multiple incoming responses are possible when using multicast with both approaches, the two new methods could once again return an `AsyncIterable` which would enable an asynchronous processing of the results.

I think both approaches (and especially the case of using DNS-SD with multicast DNS) can be motivated via use cases quite well. However, I wanted to create an issue in this regard already since it also concerns a general design question if adding new methods to the `WoT` namespace would be the way to go here or if we risk making the API surface a bit too large. Is there maybe also a concept of sub-namespaces in WebIDL?

Another question we could also discuss in this regard concerns the generic `discover` method. Should we keep it as it is right now or should we maybe also replace it with more specialized methods in the future (like `discoverNearbyThings` or `discoverThingsViaBluetooth`)? From what I've seen so far when it comes to discovery with Bluetooth and NFC, I suppose that the return type of such methods could also be a URI or an `AsyncIterable` of URIs, right?

danielpeintner:
> Another question we could also discuss in this regard concerns the generic `discover` method. Should we keep it as it is right now or should we maybe also replace it with more specialized methods in the future (like `discoverNearbyThings` or `discoverThingsViaBluetooth`)?

If we want to introduce more specialized methods I would rather use a configuration option that can be passed in in some way so that if we want to add yet another way of discovering the option needs to be updated and not yet a new method needs to be created.

JKRhb:
> If we want to introduce more specialized methods I would rather use a configuration option that can be passed in in some way so that if we want to add yet another way of discovering the option needs to be updated and not yet a new method needs to be created.

[...]

PR #545, opened 2024-02-05
feat: expand exploreDirectory algorithm

JKRhb:
This PR contains a rough first draft for expanding the `exploreDirectory` algorithm.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-scripting-api/pull/545.html" title="Last updated on Jul 8, 2024, 9:03 AM UTC (dc81bc5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/545/352fe0c...JKRhb:dc81bc5.html" title="Last updated on Jul 8, 2024, 9:03 AM UTC (dc81bc5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -754,33 +754,40 @@
           Set |discovery|.{{ThingDiscoveryProcess/[[filter]]}} to |filter:ThingFilter|.
         </li>
         <li>
-          Request the underlying platform to start the directory discovery process.
           <p class="note">
             This is a placeholder for more details in the discovery algorithm.
             Implementations should follow the procedures described in the
             [[wot-discovery]] and [wot-binding-templates] specifications.
-            Some normative steps are indicated below.
+            Some normative steps (that are currently Work in Progress) are
+            indicated below.
           </p>
-          <ol>
-            <li>
-              If |url| is not a <a>TD Directory</a> or if the underlying
-              implementation cannot support the <a>Protocol Binding</a>
-              indicated by |url|, [=reject=] |promise|
-              with {{NotSupportedError}} and terminate these steps.
-            </li>
-            <li>
-              If filters in general are not supported by the implementation and
-              |filter| is not `undefined` or `null`, [=reject=] |promise|
-              with {{NotSupportedError}} and stop.
-            </li>
-            <li>
-              Run the <a>discovery process</a> given |discovery|.
-              <p class="note">
-                From this point on, errors are recorded only on
-                {{ThingDiscoveryProcess/error}}, but don't affect |promise| any longer.
-              </p>
-            </li>
-          </ol>
+
+          If the underlying implementation cannot support the
+          <a>Protocol Binding</a> indicated by |url|, [=reject=] |promise| with
+          {{NotSupportedError}} and stop.
+        </li>
+        <li>
+          Request the underlying platform to retrieve the TD available under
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -790,33 +790,36 @@ <h3>Validating an ExposedThingInit</h3>
...
+          Let |consumedTddThing| be the result of
+          <a href="#creating-consumedthing">creating a <dfn>ConsumedThing</dfn></a>
+          from |directoryTd|.
+          If that fails, resolve |promise| with {{SyntaxError}} and stop.
+        </li>
+        <li>
+          Run the <a>discovery process</a> given |discovery| by
+          either invoking one of the actions associated with the Search API
+          – if the {{ThingDiscoveryProcess/[[filter]]}} argument is not
+          `undefined` or `null` – or reading the `things` property.
```
zolkis, line 819: This should be expanded into more concise sub-steps.

zolkis (changes requested):

index.html
JKRhb: Started the split into sub-steps, however, there is still some more refinement needed.

zolkis:
The algorithm drafted here looks more like an internal implementation detail. It is hard to standardize it in order to expose these details to scripts. Or, in other words, it's hard to distillate simple APIs that cover the programmatic side of discovery, without going too much into protocol details -- since then a script could just use the network/REST API.

I am thinking if we could make examples on how to fetch TD directories using the Discovery spec / network API, as depicted [here](https://www.w3.org/TR/wot-discovery/#discovery-overview). Those will help figuring out if we can distillate a good programmatic API on top of that.

Currently the WoT Discovery introduction mechanisms are definitely encapsulated in the Scripting implementation.

Moreover, the Authentication layer and the Exploration layer, too, are encapsulated. The Scripting API presents another (programmatic) exploration API that connects to the middleman (the Scripting implementation), provisioned as an endpoint in the given WoT solution, and maybe represented (in the future) as an internal slot to the implementation (the WoT API object). True, we are missing a standard provisioning API, but more about this later.

Since currently we are not co-hosting multiple solutions in a single WoT runtime (Scripting implementation, i.e. WoT API object), this works quite well, separating the complexities of the underlying WoT Discovery spec.

We can specify a separate API for implementing the details of the WoT Discovery spec, and that should be in a different conformance class, with its own Security section, same level as WoT Provisioning. In my view, these belong to the next level down in the stack. If we were to standardize these (Provisioning, Discovery introduction + exploration), I would support it, but eventually in a separate deliverable (spec, impl + tests) in the WG. 

If we were to co-host and support multiple WoT/IoT solutions with a single runtime and WoT object, then it needs to be transparent from identification/authentication point of view as well, therefore needing more low level configuration functions -- therefore going one level down as mentioned above. 

Another way to do that is to expose multiple WoT objects for multiple solutions, i.e. an additional API to create (async) a WoT API object, with the given provisioning/initialization routines, including discovery introduction and exploration, resulting in a configured discovery "middleman" internal slot. Once that is done, we obtain the WoT object which exposes the current API functionality.

I can draft an experimental API for this in a separate issue.

JKRhb:
Thank you, @zolkis, for your comment and the proposal in #558! I think we can probably close this PR then.

Should we maybe also close #535 since the discussion is going in a different direction now?

PR #552, opened 2024-05-15
chore: use references from specref instead local bibliography

JKRhb:
I noticed that there are currently a couple of manually defined references, which we can replace with ones provided by Specref for more consistency and – in the case of specifications that are less “stable” – automated updates whenever a new document is being published.

Note that I used the latest REC versions for TD, Discovery, and Architecture – not sure if it would make sense to also point to the latest version here instead when possible.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-scripting-api/pull/552.html" title="Last updated on May 15, 2024, 8:08 AM UTC (877153c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/552/23dc59e...JKRhb:877153c.html" title="Last updated on May 15, 2024, 8:08 AM UTC (877153c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -55,12 +55,6 @@
                "ecmascript", "streams",
                "wot-architecture11", "wot-thing-description11", "wot-binding-templates"],
         localBiblio: {
-          "WOT-DISCOVERY" : {
-            href:"https://www.w3.org/TR/2023/CR-wot-discovery-20230119/",
-            title: "Web of Things (WoT) Discovery",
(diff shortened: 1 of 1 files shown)


relu91 (approved):
Given the fact that we are lacking behind Thing Description spec, I think it might make sense to use Rec version now.

danielpeintner (approved):

Issue #555, opened 2024-06-03
Extended return type of invokeAction()


danielpeintner:
Based on the discussion in https://github.com/eclipse-thingweb/node-wot/pull/1279 two points/issues arose. One is the following.

---

At the moment the  `invokeAction(DOMString actionName, ...)` call may return `InteractionOutput` only. Doing so makes it difficult to properly handle long running actions that MAY report first an "ActionStatus" object that allows for cancelling/querying action.

The idea/proposal is to enhance the return type. Something like

`Promise<InteractionOutput | ActionStatus > invokeAction(DOMString actionName, ...) `

We need to define what `ActionStatus` might be and what can be supported.

I wonder whether we can define/collect fist ideas how `ActionStatus` could look like and we could also start experimenting with it implementation- wise. Is it possible to "label" such a feature "Experimental" so that it is clear it might change?

Note: The [EXI Profile has also an ActionStatus object](https://w3c.github.io/wot-profile/#ActionStatus) and we should make it clear that it is something else. Maybe we might also use another name...

realtes to https://github.com/w3c/wot-scripting-api/issues/335

zolkis:
That would not be a good design - how do you write use code, to use both?
Actions should then return a control object that can control the action, including intermediary commands (state, pause, resume, stop, ...), final state/value reporting etc.

lu-zero:
we have this cascade:
- invokeAction may return a success or a failure
- success may be the affordance output (if described) or any of the additionalResponse of the kind `success`
- failure may be any additionalResponse of kind `failure`
- it could be an unexpected lump of data marked as success or failure according to the protocol

it can be modeled as a `Result<AnySuccess, AnyFailure>` and then the user has to see which kind of failure or success to expect or `Result<AffordanceOutput, Result<AnySuccess, AnyFailure>>` if we assume the `AffordanceOutput` is what we care about the most.

Given that `ActionStatus` will return the `InteractionOutput` one might just always get an ActionStatus, but it is an open problem what to do with additionalResponses of success kind.

zolkis:
To me it seems it's not a plain response path, you may have business logic dependent on the return patterns, therefore it is best exposed as an interface of its own, serving the results/errors.

danielpeintner:
To continue the discussion I would like to summarize what has been discussed so far.
Based on code we are talking about something like this, right?

```
Promise<ActionControl> invokeAction(DOMString actionName, ...) 

interface ActionControl {
[...]

PR #561, opened 2024-08-26
Extended return type of invokeAction()

danielpeintner:
I started a PR to show _a_ possible solution how the extended return type `ActionInteractionOutput` for `invokeAction()` may look like.

resolves https://github.com/w3c/wot-scripting-api/issues/555

### DIFF
--- a/index.html
+++ b/index.html
@@ -1275,7 +1275,71 @@
       </ol>
     </section>
   </section>
-  <section> <h3>Using {{InteractionInput}} and {{InteractionOutput}}</h3>
+  
+  <section data-dfn-for="ActionInteractionOutput">
+    <h2>The <dfn>ActionInteractionOutput</dfn> interface</h2>
+    <p>
+      Belongs to the <a>WoT Consumer</a> conformance class.
+      An {{ActionInteractionOutput}} object is always created by the implementations
+      and exposes the data returned from <a>WoT Interactions</a> to application
+      scripts.
+    </p>
+    <p>
+      This interface exposes an action status object. Its implementation
+      will allow cancelling asynchronous actions and report the status of a long running action.
+    </p>
+    <pre class="idl">
+      enum ActionStatus {
+        /* "pending", Profile has pending. Not sure what it would mean? */
+        "running",
+        "success", /* Profile uses completed? */
+        "error" /* Profile uses failed? */
+      };
+      
+      [SecureContext, Exposed=(Window,Worker)]
+      interface ActionInteractionOutput : InteractionOutput {
+        readonly attribute object? error;
+        Promise&lt;ActionStatus&gt; status();
+        Promise&lt;undefined&gt; cancel();
+      };
+    </pre>
+    <p>
+      The <dfn>error</dfn> property represents a possible error, initially `null`.
+    </p>
+    <p class="ednote" title="Should state be a function or an attribute only?">
+      
+    </p>
+    <p class="ednote" title="Additional action object functions needed?">
+      Should we allow pause/resume also? TD has no notion of it.
+    </p>
+    <section><h3>The <dfn>status()</dfn> function</h3>
+      Reports the status of an <a>Action</a> (one of <code>running</code>, <code>success</code> or <code>error</code>) or rejects on error. The method MUST run the following steps:
+      <ol>
+        <li>
+          Return a {{Promise}} |promise:Promise| and execute the next steps
+          [=in parallel=].
+        </li>
+        <li>
+          TODO
+        </li>
+      </ol>
+    </section>
+    <section><h3>The <dfn>cancel()</dfn> function</h3>
+      Cancels a running WoT <a>Action</a> or rejects on error. The method MUST run the following steps:
+      <ol>
+        <li>
+          Return a {{Promise}} |promise:Promise| and execute the next steps
+          [=in parallel=].
+        </li>
+		<li>
+          TODO
+        </li>
+      </ol>
+    </section>
+
+  </section>
+  
+  <section> <h3>Using {{InteractionInput}}, {{InteractionOutput}} and {{ActionInteractionOutput}}</h3>
     <p>
       As illustrated in the next pictures, the {{InteractionOutput}} interface
       is used every time implementations provide data to scripts, while
@@ -1323,7 +1387,7 @@
     <p>
       When a {{ConsumedThing}} invokes an <a>Action</a>, it provides the
       parameters as {{InteractionInput}} and receives the output of the
-      <a>Action</a> as an {{InteractionOutput}} object.
+      <a>Action</a> as an {{ActionInteractionOutput}} object.
     </p>
     <p>
       An {{ExposedThing}} <a href="#the-actionhandler-callback">
@@ -1386,7 +1450,7 @@
         /*Promise&lt;undefined&gt; writeAllProperties(
                                     PropertyWriteMap valueMap,
                                     optional InteractionOptions options = {});*/
-        Promise&lt;InteractionOutput&gt; invokeAction(DOMString actionName,
+        Promise&lt;ActionInteractionOutput&gt; invokeAction(DOMString actionName,
                                     optional InteractionInput params = {},
                                     optional InteractionOptions options = {});
         Promise&lt;Subscription&gt; observeProperty(DOMString name,
--- a/typescript/scripting-api/index.d.ts
+++ b/typescript/scripting-api/index.d.ts
@@ -88,6 +88,14 @@ declare namespace WoT {
         value(): Promise<DataSchemaValue>;
     }
 
+    export enum ActionStatus { "running", "success", "error" }
+
+    export interface ActionInteractionOutput extends InteractionOutput {
+        error?: Error;
+        status(): Promise<ActionStatus>
+        cancel(): Promise<void>
+    }
+
     export interface Subscription {
         active:boolean,
         stop(options?: InteractionOptions):Promise<void>
@@ -140,9 +148,9 @@ declare namespace WoT {
          * Makes a request for invoking an Action and return the result.
          * Takes as arguments actionName, optionally params and optionally options.
          * It returns a Promise that resolves with the result of the Action represented
-         * as an InteractionOutput object, or rejects with an error.
+         * as an ActionInteractionOutput object, or rejects with an error.
          */
-        invokeAction(actionName: string, params?: InteractionInput, options?: InteractionOptions): Promise<undefined | InteractionOutput>;
+        invokeAction(actionName: string, params?: InteractionInput, options?: InteractionOptions): Promise<undefined | ActionInteractionOutput>;
 
         /**
          * Makes a request for Property value change notifications.


index.html
```diff
@@ -1275,7 +1275,71 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
+      and exposes the data returned from <a>WoT Interactions</a> to application
+      scripts.
+    </p>
+    <p>
+      This interface exposes an action status object. Its implementation
+      will allow cancelling asynchronous actions and report the status of a long running action.
+    </p>
+    <pre class="idl">
+      enum ActionStatus {
+        /* "pending", Profile has pending. Not sure what it would mean? */
```
danielpeintner, line 1293: I don't really see the reason for having it. Do you?

index.html
```diff
@@ -1275,7 +1275,71 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
+      interface ActionInteractionOutput : InteractionOutput {
+        readonly attribute object? error;
+        Promise&lt;ActionStatus&gt; status();
+        Promise&lt;undefined&gt; cancel();
+      };
+    </pre>
+    <p>
+      The <dfn>error</dfn> property represents a possible error, initially `null`.
+    </p>
+    <p class="ednote" title="Should state be a function or an attribute only?">
```
danielpeintner, line 1309: I started with a function but I start thinking an attribute should be sufficient..

index.html
zolkis: No, since the Promise should fail if the operation could not be started.

index.html
zolkis: An action is supposed to be always polled? Or could a script subscribe just for changes in status?

index.html
danielpeintner: Thanks for the feedback. If I don't get any push-back the next days I will remove the option (it was added as a comment only anyway)

index.html
```diff
@@ -1275,7 +1275,71 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
+    <p>
+      This interface exposes an action status object. Its implementation
+      will allow cancelling asynchronous actions and report the status of a long running action.
+    </p>
+    <pre class="idl">
+      enum ActionStatus {
+        /* "pending", Profile has pending. Not sure what it would mean? */
+        "running",
+        "success", /* Profile uses completed? */
+        "error" /* Profile uses failed? */
```
JKRhb, line 1296: Should we coordinate in some way with the Profile TF to align the terms that are being used? Or is it no problem if they differ, since they are operating on a different level anyway?

danielpeintner: Good point. Once we feel comfortable with our choices we should definitely reach out. Most of the status terms are 1:1 matches. What seems different is `pending`.

index.html
danielpeintner: done in https://github.com/w3c/wot-scripting-api/pull/561/commits/bb69456d1a11b2d3b3c26540d75cda9867fca39b

index.html
danielpeintner: FYI: Profile has 4 terms
* `pending` --> we don't have a corresponding term!
* `running `-> same ✅
* `completed `-> we use `success` instead
* `failed `-> we use `error` instead

What can we do with aligning the terms? Use the ones from profile or talk to them? Any opinion?

danielpeintner:
We have some inline discussions which need some more feedback. Please chime in with your opinion!

index.html
JKRhb: > Mhh, I tend to think now that a a function for reporting the status would be more powerful and useful. What do people think? Other ideas?

Hmm, if I see it correctly, we _could_ also treat the attribute as a getter and add more logic to that if needed, right?

> Or could a script subscribe just for changes in status?

If we wanted to that, what would be the best way to model that? Adding a way for registering a callback function?

index.html
JKRhb: Hmm, their terminology actually sounds quite good to me, so from my side we could also adopt what they are using.

index.html
danielpeintner: Mhh, maybe we should decide what we need/want from the status

* is it likely that requesting the status may take a while? (or put differently do we need a Promise)
* do we need a way to observe a status change (would be nice but adds complexity)

My take is that having a promise function for `status()` is fine while I am not 100% sure we should allow for listening status changes...

Are there opinions/proposals?

index.html
danielpeintner: I am fine withe the Profile terms also. The question remains if it is fine dropping "pending" then... (same same but different 🙃)
Other opinions?

danielpeintner:
@JKRhb seems to be fine with using the terms used by the profile.
@zolkis you were proposing the current terms. Any opinions?

danielpeintner:
Scripting Call 2024/12/11
Note: I try to summarize the decision which were made in the call yesterday.

* we use the terms defined by the WoT profile (`"pending", "running", "completed", "failed" `)
* we need a 5th parameter to indicate that an action has been _cancelled_
  -  -->  `"pending", "running", "completed", "failed", "cancelled"`
  - Question: shall we use `cancelled` or `canceled` (the difference seems to be British English vs American English)
* No need for `error` attribute (the status provides already a mean to indicate an error)
* we use [EventEmitter](https://nodejs.org/en/learn/asynchronous-work/the-nodejs-event-emitter) for each event
  - Q: How to model EventEmitters in WebIDL? @zolkis may be able to help (I found [this](https://github.com/intel/iot-js-api/blob/master/ocf/webidl.md))
  - there was the alternative mentioned that we could have _one_ event like `change` that reports the actual type of the change
  - what I envision is that the new interface looks somehow like this (I think that should work in TypeScript) 
```JS
   interface ActionInteractionOutput extends InteractionOutput, EventEmitter {
        // cancel a pending/running action
        cancel(): Promise<void>
        // possible events
        'pending': () => void;
        'running': () => void;
        'completed': () => void;
        'failed': (error: Error) => void;
        'cancelled': () => void;
        // Note: value() function will only work after completed state has been reached
    }
```

Please comment!
If there are no further comments/changes, I can try to implement it in this PR.

zolkis:
Nowadays the use of Promises is encouraged instead of events wherever possible (which tend to be misused). Promise are also easier to use from client code. I liked more the original design with status query returning a Promise. The latest status terms look good.

zolkis:
Cancelling operations [should be done using AbortSignal](https://www.w3.org/TR/design-principles/#aborting), but in our case a dedicated function is fine, since it's scope is strictly the referred action.

danielpeintner:
@zolkis  Thanks for your feedback!

> Nowadays the use of Promises is encouraged instead of events wherever possible (which tend to be misused). Promise are also easier to use from client code. I liked more the original design with status query returning a Promise.

I may be wrong in my thinking, but I think from a developer point of use, events seem simpler.
I can do something like the following

```
action.on('completed', () => console.log("Action completed, inform user"));
action.on('failed', (err) => console.log("Error happened: " + err));
```

Note: The only thing I am not sure about is whether we can miss events. Why am I saying that.
```
let action = await thing.invokeAction("doFoo"));
// register handlers (could it be that action is done before?)
action.on('completed', ...);
```
I think it might be possible that the emit() call comes before Í registered the handlers? 

Anyhow, in the case of a status() promise I need to pull regularly, don't I?

```
do {
  let status = await action.status();
  if (status === "failed") {
     console.log("Action failed"));
  } else if (status === "completed" ) {
     console.log("Action completed"));
  } else {
     console.log("Action not yet done..."));
  }
} while(...);
```

I don't think there is a easy way to register for an event? 
Maybe I am just not seeing the solution 🙈

@relu91 @JKRhb any opinion?

zolkis:
We won't miss events, JS is single threaded, the event loop will check for events and dispatch.

In my latest experience, devs preferred Promises to events, no setup is needed, and most interactions can be modeled with one-off transactions.

What are the main dev use cases with this API?
- canceling action is one - needs a function 
- display completion %: in this use case the client wants to control the loop and update frequency. That is easier than reacting to (and skipping some of the) events, and the loop is not spared, either. Events also need setting up. Impl already set up everything when the action started, the spec should just expose that in the control object returned by invoking the action. So I think polling the status for % is more efficient (doesn't even need Promise, just the last know status as an attribute). See also https://www.w3.org/TR/design-principles/#synchronous 
- know when the action has completed (with success or error): since it's a one-off, a Promise covers it. 
But we can experiment with implementing both, and see how they fulfill the dev use cases.

index.html
```diff
@@ -1275,7 +1275,70 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
   </section>
-  <section> <h3>Using {{InteractionInput}} and {{InteractionOutput}}</h3>
+  
+  <section data-dfn-for="ActionInteractionOutput">
+    <h2>The <dfn>ActionInteractionOutput</dfn> interface</h2>
+    <p>
+      Belongs to the <a>WoT Consumer</a> conformance class.
+      An {{ActionInteractionOutput}} object is always created by a consumer implementation
+      and exposes functionality to interact with long running (asynchronous) actions.<br />
+      Note: a synchronous action MAY not offer any <emph>additional</emph> sensible interaction beyond the ones from {{InteractionOutput}}.
```
TallTed, line 1285: - This `**MAY** not offer` is likely to be confused with `**MAY NOT** offer`, despite formatting, casing, etc. 
- Further, I'm not sure what `sensible` is contributing here. 
- I'm also left wondering whether there should be some indication of what other `interactions` (`sensible` or not) *might* be offered (if the `action` does offer more than those in `{{InteractionOutput}}`)?

Perhaps this rephrasing is acceptable, for the first two points above? Something additional might be needed for the last.
```suggestion
      Note: A synchronous action MAY offer only the interactions found in {{InteractionOutput}} and no others.
```

index.html
```diff
@@ -1275,7 +1275,70 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
...
+      The <dfn>error</dfn> property represents a possible error, initially `null`.
+    </p>
+    <p class="ednote" title="Should state be a function or an attribute only?">
+      
+    </p>
+    <p class="ednote" title="Additional action object functions needed?">
+      Should we allow pause/resume also? TD has no notion of it.
+    </p>
+    <section><h3>The <dfn>status()</dfn> function</h3>
+      Reports the status of an <a>Action</a> (one of <code>running</code>, <code>success</code> or <code>error</code>) or rejects on error. The method MUST run the following steps:
```
TallTed, line 1315:
```suggestion
      Reports the status of an <a>Action</a> (one of <code>running</code>, <code>success</code>, or <code>error</code>), or rejects on error. The method MUST run the following steps:
```

index.html
JKRhb: I've tried to cover the third aspect that you've mentioned by rephrasing the sentence as follows – hope that it is a bit clearer than it was before:

```suggestion
      Note: The output of a synchronous action MAY be limited to the functionality of a regular {{InteractionOutput}} object, e.g., invoking the `cancel()` method might have an effect.
```

TallTed: I'm not against it, if its accurate, but this revision (which would cover all of my bullets) seems to have a very different meaning than the original.

relu91:
08/1:
- the group decided that to move on here we need a definition of the result of the `queryaction` operation which unfortunately is underspecified in the TD spec. We should open an issue in the TD (or link an existing one).

relu91:
Opened https://github.com/w3c/wot-thing-description/issues/2068

TallTed:
> 08/1:

I strongly advise using [ISO8601 date and datetime and time strings](https://xkcd.com/1179/), e.g., `2025-01-08` instead of the ambiguous `08/1`, `08/01`, `01/08/, etc., for such notes.

relu91:
> I strongly advise using [ISO8601 date and datetime and time strings](https://xkcd.com/1179/), e.g., 2025-01-08 instead of the ambiguous 08/1, 08/01, `01/08/, etc., for such notes.

Yes, thanks. 

About the API for `invokeAction`, I think that given the discussion happening in https://github.com/w3c/wot-thing-description/issues/2068, I would like to see just an additional method called queryaction (or status as you wish) that invokes the underlying operation and return a InteractionOutput. 

Given that the spec is not clear about which schema to use for this result the `InteractionOutput` should:
- correctly resolve for `arrayBuffer` method and should have a valid `ReadableStream`. This allows the application to deal with payload parsing and validation which is unknown and completely dependent by protocol or application logic. 
- `value` function should throw ... the other options that I've heard here (trying to use the output schema or using addtionalResponses) feel too much workarounds.

relu91:
- @relu91 given the current design of Scripting API and the fact that the underlying issue (https://github.com/w3c/wot-thing-description/issues/2068) is going to be fixed in TD 2.0, we should strive for a simple fix. This fix can add two functions to the API interface: `cancelAction` and `queryAction`. A more complex design like the one provided here can be introduced in a new Scripting API version that is compatible with TD next. 
- @danielpeintner I would prefer the approach described in this PR but I want to see how this approach would look like and decide later
- @JKRhb I'm fine with the direction proposed
- @ashimura I'm ok to explore the addition of the new methods because the Scripting API is a note and we don't have to update the implementation report. 

The task force decided to keep this PR on hold and open a new proposal with the two functions. Any objections @zolkis?

zolkis:
The shape of the Scripting API should allow non-breaking future compatibility. 
The design in this PR (extending InteractionOutput to ActionInteractionOutput) -- or any control object -- is better, because extending it is easier and more contained than adding/cluttering the ConsumedThing interface.

Another argument for the control object pattern is implicit identification of actions. When you add `cancelAction()` and `queryAction()` to `ConsumedThing`, you'd need to add a mechanism to explicitly identify an action -- which most probably will be an object anyway, therefore we'd get both the complexity and the clutter. We could at least remove the clutter.

A third argument for the design in this PR is that actions can also represent transactional logic. Whether canceling an action is also unrolling is Thing-specific (so not necessarily always true), but the transactional support must be allowed. Here again, a control object allows encapsulating better this use case.

A fourth argument for this design is that we already use a similar pattern with the `Subscription` object.

So I have enumerated 4 arguments for going with the control object pattern, like in this PR. If the scope of the changes for now are adding `cancelAction()` and `queryAction()`, in my opinion we should add them to the action control object, not directly to `ConsumedThing`. Naming could be discussed, of course.

danielpeintner:
I would like to bring in yet another possible solution to the problem 🙈

It seems that at the moment we are struggling to find a good way to cancel/query an action given that the TD spec is underspecified. Anyhow, we currently try to find a solution which in the best case _shifts_ the actual work to the application layer (which may know more how to properly interact with an invoked action).

Having said that, I wonder a bit whether we should acknowledge the fact that we cannot do much right now. We could describe the situation in our document and "what" applications could do (if they know more about the use-case than the TD is capable to describe).
This would postpone the work in Scripting till TD2.0 has a proper language to describe *manageable* actions.

relu91:
> It seems that at the moment we are struggling to find a good way to cancel/query an action given that the TD spec is underspecified. Anyhow, we currently try to find a solution which in the best case shifts the actual work to the application layer (which may know more how to properly interact with an invoked action).

This aspect is important as general guidance for our work in scripting API. We should try our best to abstract but if we don't have enough information from TD spec we should give the tools to let the application layer build its abstraction. 

As for the issue at hand, we can use the current design (@zolkis brought up good points in favor of it). I'm just afraid that it would give the impression that the user can create multiple instances of the same running action (e.g., invoking the action twice and keeping track of the returned instance), but in the end, we can only support canceling one action (unless to use query variables, which the application needs to pass manually). Therefore, we should put a big warning and explainer text in the specification. Whereas with the other design, the user is more aware of this shortcoming because he has manually put the action information, although less future proof.  

So as a concrete step forward, I suggest finishing up this PR with the recommendation described in the paragraph above.

zolkis:
Sounds good at first look.

I am just thinking about `query` vs `cancel`: the former is the generic way, with op param, where cancel will be one of the op's.

As things get more and more concrete, i.e. op's are getting added, the web API way is to add functions to the known op's, so we'd have a `cancel` function. It makes algorithms better separated and defined. So `cancel` becomes one op exposed as a function, that may or may not be supported by the underlying platform, but its use case is clear and it can fail (for non-support or security or nondescript reasons).
If we don't know how to spec a `cancel` algorithm, we should omit it. 
Anyway, let's not call it "renaming cancel to query", but model cancel with query, for time being, until it gets specified in the TD spec.

Then, just as a note to myself, I don't like low level interfaces like exposing `formIndex` and `uriVariables`, as it's a weak abstraction to present a generic RESTful API as a JS API, but it's less friction for implementations and apps.

danielpeintner:
> I am just thinking about `query` vs `cancel`: the former is the generic way, with op param, where cancel will be one of the op's.

I am sorry, I made a mistake. I meant

> renamed `status` to `query` to align with op values

I fixed it in the original comment.

danielpeintner:
Based on today's discussion the following changes have been made:
* remove `error`  (the calls can throw errors)
* added `InteractionInput` to `query()`/`cancel()` functions

relu91:
Call 14/04/2025:
- We can move on with the prose (specification text). 
- One open point is what to do with the new functions `query` and `cancel` in corner cases like: action is sync. We can always throw in this case or simulate a behaviour (cancel -> no op, you can't cancel a sync action, query -> no op, but interaction output empty). 
- @zolkis  do nothing is bad practice, we should explain this and let the implementation decide.
- Do we have something similar in the Web space?

We would like to hear more from the group @zolkis @JKRhb, any others?

danielpeintner:
I pushed updates and opted for reporting an error if the functions cancel/query are not supported (e.g., synchronous action)

typescript/scripting-api/index.d.ts
```diff
@@ -93,7 +93,7 @@ declare namespace WoT {
         query(params?: InteractionInput, options?: InteractionOptions): Promise<InteractionOutput>
         // cancel a pending/running action
         cancel(params?: InteractionInput, options?: InteractionOptions): Promise<void>
-        // Note: retrieving the result of the action via implicit InteractionOutput interface will only work after the action has been completed
+        // Note: retrieving the result of an action via the implicit InteractionOutput interface will only work after the action has been completed
```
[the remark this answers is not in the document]
danielpeintner: You are right, commited https://github.com/w3c/wot-scripting-api/pull/561/commits/dee71ab4740bcb9742c8b363afe5f23690ec10fd

relu91 (approved):
I reviewed the changes. I think it is good enough to start experimenting. We can remove the waitForTD label, but let's discuss it today.

zolkis (approved):

relu91:
Given 3 approvals, I'm merging this PR asynchronously (an ironic twist of fate 😜).

danielpeintner:
FYI: I created https://github.com/w3c/wot-scripting-api/pull/576 so that we can experiment with ... for example in node-wot

PR #575, opened 2025-05-14
Update README with tag

danielpeintner:
`--tag latest` is necessary to show up as _latest_ on NPM


JKRhb (approved):

relu91 (approved):
Short context: this is necessary after the latest npm major release (11.x.y)

relu91:
We have two approvals, and it is not an editor change. I'm merging.

Issue #560, opened 2024-08-06
Cover AdditionalExpectedResponses as error responses


relu91:
Current TD specification introduced the concept of additionalResponses to model APIs that might return different payloads from the same affordance. This issue focuses on [AdditionalExpectedResponses](https://www.w3.org/TR/wot-thing-description11/#additionalexpectedresponse) marked with the field `success: false` (default). The goal is to understand how to treat these responses in our runtime. The discussion is strongly connected with #200.

relu91:
Call 11/12:
- @relu91 As a draft idea we can just instruct runtimes to throw one general [OperationError](https://webidl.spec.whatwg.org/#operationerror) 
- Should we throw or return the error? 
- @JKRhb Introduce a custom error type to wrap the error and access it with a value function.
- @ashimura we should talk about error handling in general in overall Scripting API, depending on potential responses.

PR #589, opened 2026-04-15
feat: handle interaction error response algorithm

relu91:
Initial draft of the algorithm to handle interaction error response, as described in issue #560. 

Discussions points:
- Introduced the concept of `interaction method` to indicate one of the methods of Consumed thing that implements a TD operation. 
- The mapping between `AdditionalExpectedResponse` and the actual response is left underspicified to let the protocol binding implementation decide (for example, the HTTP binding might use `htv:statusCodeValue` to match the right expected response)
- I introduced a dedicated custom error for error responses that match the one declared in the TD. OperationError in the end was too generic, and it didn't allow attaching the response as `InteractionOutput` as I did with `InteractionError`
- Given the fact that `InteractionOuput` take as input a Form we should override its content-type with the one declared in the `AdditionalExpectedResponse`. This change is not yet in this PR.

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/589.html" title="Last updated on May 13, 2026, 1:28 PM UTC (0b8567d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/589/10f56f0...relu91:0b8567d.html" title="Last updated on May 13, 2026, 1:28 PM UTC (0b8567d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1437,6 +1437,10 @@
 
     <section data-dfn-for="InteractionError">
       <h3>The <dfn>InteractionError</dfn> interface</h3>
+      <p class="note">
+        Currently, we are looking for wider implementation of the following section. 
+        The alorithm presented here might be subject to changes in the future. 
+      </p>
       <p>
         Represents an error that occurred during a <a>WoT Interaction</a> with
         additional diagnostic information from the <a>Thing</a> in the form of
@@ -1518,16 +1522,6 @@
         </li>
       </ol>
     </section>
-
-    <p class="ednote">
-      This topic is still being discussed in
-      <a href="https://github.com/w3c/wot-scripting-api/issues/200">Issue #200</a>.
-      A standardized error mapping would be needed in order to ensure consistency
-      in mapping script errors to protocol errors and vice versa. In particular,
-      when algorithms say "error received from the <a>Protocol Bindings</a>",
-      that will be factored out as an explicit error mapping algorithm. Currently,
-      that is encapsulated by implementations.
-    </p>
   </section>
 
   </section> <!-- Handling interaction data -->
--- a/index.html
+++ b/index.html
@@ -1779,10 +1779,25 @@
             Let |result:object| be an object and for each string |name:string| in |propertyNames| add a property with key |name| and the value `null`.
           </li>
           <li>
+            If this cannot be done with a single request with the <a>Protocol Bindings</a>, [=reject=] |promise| with a
+            {{NotSupportedError}} and stop.
+          </li>
+          <li>
             Make a request to the underlying platform (via the <a>Protocol Bindings</a>) to retrieve the <a>Property</a> values given by |propertyNames| with |form| and optional URI templates given in |options|' |uriVariables|.
           </li>
           <li>
-            If this cannot be done with a single request with the <a>Protocol Bindings</a>, [=reject=] |promise| with a {{NotSupportedError}} and stop.
+            If the request fails with error |receivedError| and optionally |failedResponse|,
+            run the following sub-steps:
+            <ol>
+              <li>
+                let |error| be the result of running
+                <a>handle interaction error response</a> given the |receivedError|
+                |form| and |failedResponse|.
+              </li>
+              <li>
+                [=Reject=] |promise| with |error| and stop.
+              </li>
+            </ol>
           </li>
           <li>
             Process the response and for each |key| in |result|, run the following
@@ -1850,7 +1865,18 @@
             If this cannot be done with a single request with the <a>Protocol Bindings</a> of the <a>Thing</a>, then [=reject=] |promise| with a {{NotSupportedError}} and stop.
           </li>
           <li>
-            If the request fails, [=reject=] |promise| with the error received from the <a>Protocol Bindings</a> and stop.
+            If the request fails with error |receivedError| and optionally |failedResponse|,
+            run the following sub-steps:
+            <ol>
+              <li>
+                let |error| be the result of running
+                <a>handle interaction error response</a> given the |receivedError|
+                |form| and |failedResponse|.
+              </li>
+              <li>
+                [=Reject=] |promise| with |error| and stop.
+              </li>
+            </ol>
           </li>
           <li>
             Process the reply and let |result:object| be an object with the keys and values obtained in the reply.
@@ -1920,7 +1946,18 @@
             |options|' |uriVariables|.
           </li>
           <li>
-            If the request fails, [=reject=] |promise| with the error received from the <a>Protocol Bindings</a> and stop.
+            If the request fails with error |receivedError| and optionally |failedResponse|,
+            run the following sub-steps:
+            <ol>
+              <li>
+                let |error| be the result of running
+                <a>handle interaction error response</a> given the |receivedError|
+                |form| and |failedResponse|.
+              </li>
+              <li>
+                [=Reject=] |promise| with |error| and stop.
+              </li>
+            </ol>
           </li>
           <li>
             Otherwise [=resolve=] |promise|.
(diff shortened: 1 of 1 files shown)


danielpeintner (commented):
Some very high level comments/questions

* Do we need the concept of  "interaction method"? Why can't we use `op` right away ? I did not quite get that point 🙈
* I think the editors not right before section 8 needs to move up

relu91:
> Do we need the concept of "interaction method"? Why can't we use op right away ? I did not quite get that point

I needed the new definition to quickly reference all the ConsumedThing methods that implement the WoTInterface here:
>This algorithm is used by <a>interaction methods</a> to process error responses
        from the <a>Protocol Bindings</a> and determine whether to create an
        {{InteractionError}} or map to a standard error type.
        
 Ofc, we can remove it if you have better suggestions. 
 
 > I think the editors note right before section 8 needs to move up
 
 Yes, that is another discussion point: should we keep it?

index.html
```diff
@@ -1434,6 +1438,91 @@ <h2>The <dfn>ActionInteractionOutput</dfn> interface</h2>
...
+        failed <a>WoT Interaction</a>.
+      </p>
+    </section>
+
+    <section>
+      <h3>The <dfn>handle interaction error response</dfn> algorithm</h3>
+      <p>
+        This algorithm is used by <a>interaction methods</a> to process error responses
+        from the <a>Protocol Bindings</a> and determine whether to create an
+        {{InteractionError}} or map to a standard error type.
```
egekorkan: I was initially confused with the standard error type but read further and understood that it means a DOMException. Am I right?

danielpeintner: I think @relu91 is hinting at the other errors we are using like NotFoundError, SecurityError, NotSupportedError , SyntaxError ...

Correct?

relu91:
From today's call:
- In the past, we used https://webidl.spec.whatwg.org/#idl-DOMException-error-names as a basic reference for already defined errors. However, the semantic it is a little bit stretched as the DOMException relates to errors revolving around HTML parsing. As you can see in the document, we found some relevant mappings (NotFoundError, NotAllowedError...) that need to be refactored if we decide to use another error reference. In the future, we should review if it makes sense to continue to use this mapping. @zolkis's opinion would be valuable here. 
- Probably we should also cover the ExposedThing side when dealing with `AdditionalResponses`


I covered the comments and finished referencing the algortihm in relevant section. @danielpeintner and @egekorkan please have a look.

index.html
```diff
@@ -1434,15 +1434,94 @@ <h2>The <dfn>ActionInteractionOutput</dfn> interface</h2>
...
-      when algorithms say "error received from the <a>Protocol Bindings</a>",
-      that will be factored out as an explicit error mapping algorithm. Currently,
-      that is encapsulated by implementations.
-    </p>
+
+    <section data-dfn-for="InteractionError">
+      <h3>The <dfn>InteractionError</dfn> interface</h3>
+      <p class="note">
+        Currently, we are looking for wider implementation of the following section. 
+        The alorithm presented here might be subject to changes in the future.
```
danielpeintner:
```suggestion
        The algorithm presented here might be subject to changes in the future. 
```

index.html
```diff
@@ -1434,15 +1434,94 @@ <h2>The <dfn>ActionInteractionOutput</dfn> interface</h2>
...
+                  Let |responseSchema| be |responseEntry|.|schema|, if it exists.
+                </li>
+                <li>
+                  Let |errorOutput:InteractionOutput| be the result of running
+                  <a>parse interaction response</a> on |response|, |form|, and
+                  |responseSchema|.
+                </li>
+                <li>
+                  Return a new {{InteractionError}} constructed with |errorOutput| an appropriate
+                  message derived from the |binding| or |error|.
```
danielpeintner: This sentence reads a bit strange.
What about

> Return a new InteractionError constructed with errorOutput. Please ensure the message is appropriately derived from the specific binding or error encountered.

index.html
```diff
@@ -1678,11 +1768,26 @@ <h3>The <dfn>readMultipleProperties()</dfn> method</h3>
           <li>
             Let |result:object| be an object and for each string |name:string| in |propertyNames| add a property with key |name| and the value `null`.
           </li>
+          <li>
+            If this cannot be done with a single request with the <a>Protocol Bindings</a>, [=reject=] |promise| with a
+            {{NotSupportedError}} and stop.
+          </li>
```
danielpeintner, line 1775: Good catch 👍

index.html
```diff
@@ -2389,8 +2557,22 @@ <h4>The <dfn>stop()</dfn> method</h4>
...
+                  |unsubscribeForm| and |failedResponse|.
+                </li>
+                <li>
+                  [=Reject=] |promise| with |error| and stop.
+                </li>
+                <li>
+                  If the underlying platform receives further notifications for this
+                  subscription, implementations SHOULD silently suppress them.
+                </li>
+              </ol>
```
danielpeintner, line 2576: The only thing that bothers me a bit is that we duplicated this bullet list item X times. I wonder whether we can have a local place and point to it. Unfortunetla that would mean when going through the algorithm one needs to jump to another place...

danielpeintner (commented):
Looks very good. Some comments below.

zolkis:
> * In the past, we used https://webidl.spec.whatwg.org/#idl-DOMException-error-names as a basic reference for already defined errors. However, the semantic it is a little bit stretched as the DOMException relates to errors revolving around HTML parsing. As you can see in the document, we found some relevant mappings (NotFoundError, NotAllowedError...) that need to be refactored if we decide to use another error reference. In the future, we should review if it makes sense to continue to use this mapping. @zolkis's opinion would be valuable here.

I can only make non-substantial contributions now (as I am not a group member any more).

The mismatch comes from the fact the Scripting API is a W3C spec, and as such, uses the Infra. 
In practice, the implementations are npm based, so a W3C spec is not even needed. 
Even more since browser implementations are unlikely (close to 0 probability).
But the spec can override the exception handling, even if it's syntactically a W3C spec -- but then expect issues when someone eventually (theoretically) decides to implement the spec in the browser :) :).

index.html
danielpeintner: Unless we find something clever we have to stick to it 🙈

index.html
```diff
@@ -1434,15 +1434,94 @@ <h2>The <dfn>ActionInteractionOutput</dfn> interface</h2>
...
+      </p>
+      <p>
+        Represents an error that occurred during a <a>WoT Interaction</a> with
+        additional diagnostic information from the <a>Thing</a> in the form of
+        response data associated with an <a>AdditionalExpectedResponse</a> defined in
+        the <a>Thing Description</a>.
+      </p>
+      <pre class="idl">
+        [SecureContext, Exposed=(Window,Worker)]
+        interface InteractionError : Error {
```
relu91, line 1452: WebIDL suggests extending `DOMException`, I need to update this before merging, since the discussion (`Error` vs `DOMException` is still open.

relu91: @danielpeintner turns out that if we want to extend DOMException, the "extra data" we carry in the Exception must be `serializable`. Sadly, `InteractionOutput` is not serializable as it contains a non-serializable field ( [ReadableStream](https://streams.spec.whatwg.org/#readablestream)? [data](https://github.com/w3c/wot-scripting-api/pull/589#dom-interactionoutput-data)). We can:
- Leave the changes as they are, but we have this inconsistency between the "native" errors and our "custom" error. The inconsistency is that InteractionError does not extend `DOMException`. 
- We force the thrower of the error to eagerly read the full payload and then throw.

danielpeintner: It seems easiest to stick with Error since we are not happy with DOMException anyway.

relu91: It seems we have the same problem with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error . However, I'm aware that some HTTP clients do not respect the serializable constraint.

danielpeintner: Mhh, what are you proposing?

relu91: I will review again what `fetch` and `axios` are doing. The thing is that having lazy methods for consuming the network payload is an important feature that we should not drop.

PR #590, opened 2026-04-29
Tweak arrayBuffer() function and stop() method text slightly

danielpeintner:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/pull/590.html" title="Last updated on May 13, 2026, 9:57 AM UTC (499cb32)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/590/10f56f0...499cb32.html" title="Last updated on May 13, 2026, 9:57 AM UTC (499cb32)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1008,7 +1008,7 @@
     </section>
 
     <section><h3>The  <dfn>arrayBuffer()</dfn> function</h3>
-      When invoked, MUST run the following steps:
+      When invoked, it MUST run the following steps:
       <ol>
         <li>
           Return a {{Promise}} |promise:Promise| and execute the next steps
@@ -1021,7 +1021,7 @@
         <li>
           Let |reader| be the result of
           <a data-cite="streams#readablestream-get-a-reader">
-          getting a reader</a> from |data|. If that threw an exception, [=reject=]
+          getting a reader</a> from |data|. If that throws an exception, [=reject=]
           |promise| with that exception and stop.
         </li>
         <li>
--- a/index.html
+++ b/index.html
@@ -1007,8 +1007,8 @@
       </p>
     </section>
 
-    <section><h3>The  <dfn>arrayBuffer()</dfn> function</h3>
-      When invoked, it MUST run the following steps:
+    <section><h3>The <dfn>arrayBuffer()</dfn> function</h3>
+      When invoked, the `arrayBuffer()` function MUST run the following steps:
       <ol>
         <li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1008,7 +1008,7 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
     </section>
 
     <section><h3>The  <dfn>arrayBuffer()</dfn> function</h3>
-      When invoked, MUST run the following steps:
+      When invoked, it MUST run the following steps:
```
danielpeintner: Or maybe "the function MUST run the following steps:" ?

TallTed: Either works. I might go so far as ```the `arrayBuffer()` function MUST run the following steps```

danielpeintner: We have some similar wordings in other places, e.g., 
* https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/pull/590.html#the-stop-method

BTW, w.t.r. I noticed we use "stop() **function**" vs. "arrayBuffer() **method**".
I wonder whether we should use "method" in the case of "value()", "arrayBuffer()", "query()",  "cancel()" etc since it is associated with a class/object and not a _static_ function per se 🤔

relu91: Ok for applying the same pattern to all the occurrences:
> the `arrayBuffer()` function MUST run the following steps

danielpeintner: I made those changes, please take a look.

The alignments w.r.t. method vs function we tackle in a follow-up (see https://github.com/w3c/wot-scripting-api/issues/593)

relu91 (approved):

index.html
```diff
@@ -1007,8 +1007,8 @@ <h2>The <dfn>InteractionOutput</dfn> interface</h2>
       </p>
     </section>
 
-    <section><h3>The  <dfn>arrayBuffer()</dfn> function</h3>
-      When invoked, MUST run the following steps:
+    <section><h3>The <dfn>arrayBuffer()</dfn> function</h3>
+      When invoked, the `arrayBuffer()` function MUST run the following steps:
```
TallTed, line 1011: This appears to be the only "MUST _run_ the following", at least as visible on this page. The others are "MUST _execute_ the following". I don't have a strong feeling on this, but it seems odd to have one differ from the others, so I'm flagging it as something the editor(s) might want to change.

danielpeintner: Created https://github.com/w3c/wot-scripting-api/issues/596 to keep track

Issue #539, opened 2024-01-27
ObserveProperty with uri vars


ferllings:
Hello,
Could someone clarifies something for me?
In the ObserveProperty method: https://w3c.github.io/wot-scripting-api/#the-observeproperty-method
The #6 says
```
If thing.[activeObservations][propertyName] [=map/exists], [reject] promise with a [NotAllowedError] and stop.
```
But in the case of using uriVars, do we consider the subscription is the same, even if the uriVars are different?

#8 indicates:
```
Make a request to the underlying platform to observe the [Property] identified by propertyName with form and optional URI templates given in options' uriVariables.
```

What if we already made a previous request, `.../propertyA?v=X`
but we want to make a new request for `.../propertyA?v=Y`

Thanks for your help?

relu91:
> And I believe there is a mistake in subscribeEvent() https://w3c.github.io/wot-scripting-api/#the-subscribeevent-method
https://github.com/w3c/wot-scripting-api/issues/6 says
>
>If thing.[activeSubscriptions][eventName] DOES NOT [exist](https://infra.spec.whatwg.org/#map-exists), [reject](https://webidl.spec.whatwg.org/#reject) promise with a [NotAllowedError] and stop.

Related to #540 .

For question, yes with the current approach, you can't have two active subscriptions to the same `Event`. This means that if you need it you have two options:
1. you subscribe and then unsubscribe and then subscribe again
2. You consume the Thing twice and then use the second instance to subscribe with the new URI variables

I know that both options are ideal, we can work to fix this issue in the future release. To understand better it would be valuable if you provide a short description of the concrete use case you have ( protocol you are using, platform, etc.? ). 

@zolkis: anyhow it is a bug since we are supporting URIVariables you should consider the two requests as two different subscriptions.

ferllings:
@relu91 Thanks for the clarification
For the context, I'm running my own golang implementation of WoT, for home automation.
It only supports http/ws for now.

I have a (thing) service that provides some dynamic states for my house.
And clients that observe those states, using filters to trigger various automations.
something like /isAtHome?name=me or /isAtHome?name=wife

I'll find a workaround, until this get clarified in the next version

relu91:
Call 14/05/2025:
- @relu91, before moving on and resolving this, I will review why we restricted the number of subscriptions to one per event/property. We will discuss again in the next call

zolkis:
We should reopen this discussion with the presented use case in mind.

[...]

PR #603, opened 2026-07-22
Clarify the role of URI variables in observe and subscribe ops

relu91:
Implements the fix proposed by @danielpeintner in #539: a Property/Event can now have multiple active Subscriptions at once, as long as they use different URI variables, instead of always being limited to one.

**Changes**

- New create a subscription identifier algorithm, keyed on name + sorted/serialized uriVariables.
observeProperty()/subscribeEvent() now key [[activeObservations]]/[[activeSubscriptions]] on this identifier instead of the bare property/event name.
- Added Subscription.[[identifier]] internal slot so stop() removes the right map entry.
- Updated ed-notes and slot descriptions to match.

**Open discussion point**: the added note says implementations "MUST NOT multiplex several Subscriptions onto a single underlying protocol-level subscription." That's normative (RFC2119) language sitting inside a class="note" block, which is conventionally non-normative. Should that requirement be pulled out into a numbered algorithm step (or its own normative paragraph) instead, leaving the note purely explanatory?

***
<a href="https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/603.html" title="Last updated on Jul 22, 2026, 3:25 PM UTC (b0c6ac5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-scripting-api/603/af77f9b...relu91:b0c6ac5.html" title="Last updated on Jul 22, 2026, 3:25 PM UTC (b0c6ac5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1538,14 +1538,18 @@
           <td><dfn>[[\activeSubscriptions]]</dfn></td>
           <td>`{}`</td>
           <td>
-            An [=ordered map=] [=map/keyed=] on a [=string=] name representing the <a>Event</a> and [=map/value=] is a {{Subscription}} object.
+            An [=ordered map=] [=map/keyed=] on a <a>subscription identifier</a>
+            derived from the <a>Event</a> name, and [=map/value=] is a
+            {{Subscription}} object.
           </td>
         </tr>
         <tr>
           <td><dfn>[[\activeObservations]]</dfn></td>
           <td>`{}`</td>
           <td>
-            An [=ordered map=] [=map/keyed=] on a [=string=] name representing a <a>Property</a> and [=map/value=] is a {{Subscription}} object.
+            An [=ordered map=] [=map/keyed=] on a <a>subscription identifier</a>
+            derived from the <a>Property</a> name, and [=map/value=] is a
+            {{Subscription}} object.
           </td>
         </tr>
         </tbody>
@@ -1915,6 +1919,64 @@
     </section>
 
     <section>
+      <h3>The <dfn data-lt="subscription identifier">create a subscription identifier</dfn> algorithm</h3>
+      <div>
+        A single <a>Property</a> or <a>Event</a> can be the target of several
+        active {{Subscription}}s at the same time, as long as they are made
+        with different URI variables. To distinguish between them, a
+        <a>subscription identifier</a> is used as the key of the
+        {{ConsumedThing/[[activeObservations]]}} and
+        {{ConsumedThing/[[activeSubscriptions]]}} internal slots instead of
+        plainly using the <a>Property</a> or <a>Event</a> name.
+        <p>
+          To <a>create a subscription identifier</a> given |name:string| and
+          |options:InteractionOptions|, run the following steps:
+        </p>
+        <ol>
+          <li>
+            Let |identifier| be |name|.
+          </li>
+          <li>
+            If |options|.|uriVariables| [=map/exists=], then:
+            <ol>
+              <li>
+                Let |keys| be the result of getting the keys of
+                |options|.|uriVariables|, sorted in ascending code point
+                order.
+              </li>
+              <li>
+                For each |key| in |keys|, set |identifier| to the
+                concatenation of |identifier|, `"&"`, |key|, `"="`, and the
+                result of running <a data-cite=
+                "infra#serialize-a-javascript-value-to-a-json-string">
+                serialize a JavaScript value to a JSON string</a> on
+                |options|.|uriVariables|[|key|].
+              </li>
+            </ol>
+          </li>
+          <li>
+            Return |identifier|.
+          </li>
+        </ol>
+        <p class="note">
+          <a>Protocol Bindings</a> MAY apply their own, more elaborate,
+          rationale for deciding whether two requests represent distinct
+          subscriptions. At the time of writing, [[!wot-thing-description11]]
+          only allows a client to distinguish requests to observe a
+          <a>Property</a> by their URI variables, and requests to subscribe to
+          an <a>Event</a> by their URI variables and subscription data;
+          implementations are therefore not expected to need anything more
+          elaborate than the algorithm above. Regardless of the algorithm
+          used, an implementation MUST NOT multiplex several {{Subscription}}s
+          (e.g., several listener functions) onto a single underlying
+          protocol-level subscription: every active {{Subscription}} is
+          expected to correspond to its own request to, and its own
+          notifications from, the underlying platform.
+        </p>
+      </div>
+    </section>
+
+    <section>
       <h3>The <dfn>observeProperty()</dfn> method</h3>
       <div>
         Makes a request for <a>Property</a> value change notifications.
@@ -1922,9 +1984,12 @@
         optionally |onerror:ErrorListener| and |options:InteractionOptions|.
         It returns a {{Promise}} that resolves on success and rejects on failure.
         <p class="ednote">
-          This algorithm allows for only one active {{Subscription}} per <a>Property</a>. If a new
-          {{Subscription}} is made while an existing {{Subscription}} is active the runtime
-          will throw an {{NotAllowedError}}.
+          This algorithm allows for only one active {{Subscription}} per
+          <a>Property</a> and <a>subscription identifier</a> (i.e., per
+          combination of URI variables used to observe the <a>Property</a>).
+          If a new {{Subscription}} is made with the same <a>Property</a> and
+          the same URI variables while an existing {{Subscription}} is active,
+          the runtime will throw an {{NotAllowedError}}.
         </p>
         The method MUST run the following steps:
         <ol>
@@ -1946,7 +2011,12 @@
             with a {{TypeError}} and stop.
           </li>
           <li>
-            If |thing|.{{ConsumedThing/[[activeObservations]]}}[|propertyName|] [=map/exists=], [=reject=] |promise| with a {{NotAllowedError}} and stop.
+            Let |observationId| be the result of running the
+            <a>create a subscription identifier</a> steps given |propertyName|
+            and |options|.
+          </li>
+          <li>
+            If |thing|.{{ConsumedThing/[[activeObservations]]}}[|observationId|] [=map/exists=], [=reject=] |promise| with a {{NotAllowedError}} and stop.
           </li>
           <li>
             Let |subscription| be a new {{Subscription}} object with its <a>internal slots</a>
@@ -1957,6 +2027,9 @@
               </li>
               <li>
                 Let |subscription|.{{Subscription/[[name]]}} be |propertyName|.
+              </li>
+              <li>
+                Let |subscription|.{{Subscription/[[identifier]]}} be |observationId|.
               </li>
               <li>
                 Let |subscription|.{{Subscription/[[interaction]]}} be  {{ConsumedThing/[[td]]}}.|properties|[|propertyName|].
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1921,10 +1925,13 @@ <h3>The <dfn>observeProperty()</dfn> method</h3>
         Takes as arguments |propertyName:string|, |listener:InteractionListener| and
         optionally |onerror:ErrorListener| and |options:InteractionOptions|.
         It returns a {{Promise}} that resolves on success and rejects on failure.
-        <p class="ednote">
-          This algorithm allows for only one active {{Subscription}} per <a>Property</a>. If a new
-          {{Subscription}} is made while an existing {{Subscription}} is active the runtime
-          will throw an {{NotAllowedError}}.
+        <p class="note">
+          This algorithm allows for only one active {{Subscription}} per
+          <a>Property</a> and <a>subscription identifier</a> (e.g., per
```
danielpeintner, line 1930:
```suggestion
          <a>Property</a> or <a>subscription identifier</a> (e.g., per
```

I think it is an OR (if we keep both)

danielpeintner, line 1930:
```suggestion
          <a>subscription identifier</a> (e.g., per
```

Can't we leave out "property" since the "subscription identifier" takes the property name and all the other information?

index.html
```diff
@@ -1946,7 +1953,12 @@ <h3>The <dfn>observeProperty()</dfn> method</h3>
             with a {{TypeError}} and stop.
           </li>
           <li>
-            If |thing|.{{ConsumedThing/[[activeObservations]]}}[|propertyName|] [=map/exists=], [=reject=] |promise| with a {{NotAllowedError}} and stop.
+            Let |observationId| be the result of running the
+            <a>create a subscription identifier</a> operation given
+            |propertyName| and |options|.
```
danielpeintner, line 1958: Shall we call it "create interaction identfier" ?
The reason why I say it is that we use it for subscription (event) and observation (property). In the case of property it reads a bit strange to speak about subscription..

index.html
```diff
@@ -2469,6 +2507,46 @@ <h4>The <dfn>stop()</dfn> method</h4>
       </section>
     </section>
 
+    <section>
+      <h3><dfn data-lt="subscription identifier">Create a subscription identifier</dfn> operation</h3>
```
danielpeintner, line 2511: AS mentioned before "subscription" targets more events ... not sure if we can find a better name?
"Interaction identifier" ?

index.html
```diff
@@ -2469,6 +2507,46 @@ <h4>The <dfn>stop()</dfn> method</h4>
...
+      <div>
+        A single <a>Property</a> or <a>Event</a> can be the target of several
```
danielpeintner, line 2513:
```suggestion
        A single <a>Property</a> or <a>Event</a> can be the target by several
```

Not sure but "for" reads strange to me

danielpeintner (commented):
I think the generic approach is good.

Please find some comments below

index.html
TallTed: Not sure where you came up with "for", nor why "target by" seems correct to you? 

The language I would use is "target of". [Google concurs](https://www.google.com/search?q=%22+be+the+target+of+several%22+vs+%22+be+the+target+by+several%22).
```suggestion
        A single <a>Property</a> or <a>Event</a> can be the target of several
```

danielpeintner: I misread the whole sentence. You are right, "target of" is fine.
I close/resolve this point.

index.html
relu91: I wanted the text to clearly inform the user that the runtime will not support multiple Subscriptions (i.e., Observations) for the same property given the same """"parameters""" (those parameters being encapsulated by the subscription identifier). 

If we apply your suggestions, I have the feeling that the text is less deliberate about this constraint. Maybe @TallTed can help here?

index.html
relu91: Is not really an interaction identifier, but it identifies the [Subscription](https://www.w3.org/TR/wot-scripting-api/#the-subscription-interface), which itself contains the reference to an Event or a Property:
> Represents a subscription to [Property](https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/603.html#dfn-properties) change and [Event](https://pr-preview.s3.amazonaws.com/relu91/wot-scripting-api/pull/603.html#dfn-events) interactions.

danielpeintner: Maybe what I said was not clear. My understanding is that
* for properties, we speak about observing _while_
* for events, we speak about subscribing.

What you currently use is the term "subscription," which, in my reading, is about events only and leaves out properties. Since we want to target events **and** properties, I would like to see another term used, like "interaction identifier" or anything else that targets both.

Does that make sense?

relu91: Yes, yet they both create a **subscription** ... are you saying that we should change Let |observationId| to be |subscriptionId|? I think we had the discussion on how to factor out the result of observing and subscribing and we ended up using Subscribe for both.

danielpeintner: You are right; we use "subscription" for both events and properties. Maybe it is fine then...

index.html
```diff
@@ -1538,14 +1538,18 @@ <h4>Internal slots for {{ConsumedThing}}</h4>
           <td><dfn>[[\activeSubscriptions]]</dfn></td>
           <td>`{}`</td>
           <td>
-            An [=ordered map=] [=map/keyed=] on a [=string=] name representing the <a>Event</a> and [=map/value=] is a {{Subscription}} object.
+            An [=ordered map=] [=map/keyed=] on a <a>subscription identifier</a>
+            derived from the <a>Event</a> name, and [=map/value=] is a
```
TallTed, line 1542:
```suggestion
            derived from the <a>Event</a> name; each [=map/value=] is a
```

index.html
```diff
@@ -1538,14 +1538,18 @@ <h4>Internal slots for {{ConsumedThing}}</h4>
...
+            {{Subscription}} object.
           </td>
         </tr>
         <tr>
           <td><dfn>[[\activeObservations]]</dfn></td>
           <td>`{}`</td>
           <td>
-            An [=ordered map=] [=map/keyed=] on a [=string=] name representing a <a>Property</a> and [=map/value=] is a {{Subscription}} object.
+            An [=ordered map=] [=map/keyed=] on a <a>subscription identifier</a>
+            derived from the <a>Property</a> name, and [=map/value=] is a
```
TallTed, line 1551:
```suggestion
            derived from the <a>Property</a> name; each [=map/value=] is a
```

index.html
```diff
@@ -1921,10 +1925,13 @@ <h3>The <dfn>observeProperty()</dfn> method</h3>
...
-          This algorithm allows for only one active {{Subscription}} per <a>Property</a>. If a new
-          {{Subscription}} is made while an existing {{Subscription}} is active the runtime
-          will throw an {{NotAllowedError}}.
+        <p class="note">
+          This algorithm allows for only one active {{Subscription}} per
+          <a>Property</a> and <a>subscription identifier</a> (e.g., per
+          combination of URI variables used to observe the <a>Property</a>).
+          If a new {{Subscription}} is made with the same <a>Property</a> and
+          the same URI variables while an existing {{Subscription}} is active,
+          the runtime will throw an {{NotAllowedError}}.
```
TallTed, line 1934:
```suggestion
          This algorithm allows only one active {{Subscription}} per
          combination of <a>Property</a> and <a>subscription identifier</a>,
          where the <a>subscription identifier</a> is the combination of URI
          variables used to observe the <a>Property</a>. If there is an attempt
          to make a new {{Subscription}} with the same <a>Property</a> and the
          same <a>subscription identifier</a> while an existing {{Subscription}}
          is active, the runtime will throw a {{NotAllowedError}}.
```

TallTed: or
```suggestion
          This algorithm restricts each combination of <a>Property</a> and
          <a>subscription identifier</a> (the URI variables used to observe
          the Property) to one active {{Subscription}}. Attempting to create
          a new {{Subscription}} with an identical <a>Property</a> and
          <a>subscription identifier</a> while the current one is active
          will cause the runtime to throw a {{NotAllowedError}}.
```

index.html
TallTed: I think replacing the whole block is better than line-by-line. Two options follow...

index.html
```diff
@@ -1946,7 +1953,12 @@ <h3>The <dfn>observeProperty()</dfn> method</h3>
             with a {{TypeError}} and stop.
           </li>
           <li>
-            If |thing|.{{ConsumedThing/[[activeObservations]]}}[|propertyName|] [=map/exists=], [=reject=] |promise| with a {{NotAllowedError}} and stop.
+            Let |observationId| be the result of running the
+            <a>create a subscription identifier</a> operation given
+            |propertyName| and |options|.
+          </li>
+          <li>
+            If |thing|.{{ConsumedThing/[[activeObservations]]}}[|observationId|] [=map/exists=], [=reject=] |promise| with a {{NotAllowedError}} and stop.
```
TallTed, line 1961:
```suggestion
            If |thing|.{{ConsumedThing/[[activeObservations]]}}[|observationId|]
            [=map/exists=], [=reject=] |promise| with a {{NotAllowedError}} and
            stop.
```

index.html
```diff
@@ -2107,9 +2122,12 @@ <h3>The <dfn>subscribeEvent()</dfn> method</h3>
...
         <p class="ednote">
-          This algorithm allows for only one active {{Subscription}} per <a>Event</a>. If a new
-          {{Subscription}} is made while an existing {{Subscription}} is active the runtime
-          will throw an {{NotAllowedError}}.
+          This algorithm allows for only one active {{Subscription}} per
+          <a>Event</a> and <a>subscription identifier</a> (e.g., per
+          combination of URI variables used to subscribe to the <a>Event</a>).
+          If a new {{Subscription}} is made with the same <a>Event</a> and the
+          same URI variables while an existing {{Subscription}} is active, the
+          runtime will throw an {{NotAllowedError}}.
```
TallTed, line 2130: Needs same fix as [above, lines 1929-1935](https://github.com/w3c/wot-scripting-api/pull/603/files#r3737860281).

index.html
```diff
@@ -2469,6 +2507,46 @@ <h4>The <dfn>stop()</dfn> method</h4>
...
+      <h3><dfn data-lt="subscription identifier">Create a subscription identifier</dfn> operation</h3>
+      <div>
+        A single <a>Property</a> or <a>Event</a> can be the target of several
+        active {{Subscription}}s at the same time, as long as they can be
+        told apart. To distinguish between them, a <a>subscription
+          identifier</a> is used as the key of the
+        {{ConsumedThing/[[activeObservations]]}} and
+        {{ConsumedThing/[[activeSubscriptions]]}} internal slots instead of
+        plainly using the <a>Property</a> or <a>Event</a> name.
+        <p>
```
TallTed, line 2520:
```suggestion
        <p>
          One <a>Property</a> or <a>Event</a> can be the target of several
          active {{Subscriptions}} at the same time, as long as they can be
          distinguished. For this purpose, a <a>subscription identifier</a>
          is used as the key of the {{ConsumedThing/[[activeObservations]]}}
          and {{ConsumedThing/[[activeSubscriptions]]}} internal slots,
          instead of using the <a>Property</a> or <a>Event</a> name alone.
        </p>
        <p>
```

index.html
```diff
@@ -2469,6 +2507,46 @@ <h4>The <dfn>stop()</dfn> method</h4>
...
+          protocol binding, so that no two {{Subscription}}s ever share a single
+          protocol-level subscription.
+        </p>
+        <p class="note">
+          At the time of writing, [[!wot-thing-description11]] only lets a
+          client distinguish requests to observe a <a>Property</a> by their
+          URI variables, and requests to subscribe to an <a>Event</a> by
+          their URI variables and subscription data. Implementations are
+          therefore expected to derive the identifier from |name| together
+          with those values.
```
TallTed, line 2537:
```suggestion
          Currently, [[!wot-thing-description11]] limits request
          differentiation to URI variables for observing a <a>Property</a>,
          and URI variables plus subscription data for subscribing to an
          <a>Event</a>. Implementations must therefore derive the identifier
          from |name| combined with these values.
```

